PR #27, opened 2017-08-18
Cinematiccamera

tobiasrau:


gralkapk:
View2D connected to TimeLineRenderer should invoke "ResetView" after window resize.
If not, view scaling does not fit to new window size.

PR #44, opened 2017-09-19
NGMesh and ArchVis Plugins

invor:
NGMesh Plugin: A modern renderer for arbitrary meshes. All data (shader, mesh, draw commands, shader parameters) is prepared and supplied by the data source. An example/template for such a datasource will be included.

ArchVis Plugin: Plugin for scientific architectural visualizations in the context of SFB1244.


plugins/ng_mesh/include/ng_mesh/CallNGMeshRenderBatches.h
```diff
@@ -0,0 +1,303 @@
...
+
+	class NG_MESH_API CallNGMeshRenderBatches : public megamol::core::AbstractGetData3DCall
+	{
+		public:
+
+			class NG_MESH_API RenderBatchesData
+			{
+			public:
+
+				static uint32_t shaderUpdateBit()			{ return 0x1; }
```
reinago: why? plus it's not even inline. enums work okay, don't they?

plugins/ng_mesh/include/ng_mesh/CallNGMeshRenderBatches.h
```diff
@@ -0,0 +1,303 @@
...
+					delete raw_buffer;
+					raw_buffer = new uint8_t[allocated_byte_size];
+					std::memcpy(raw_buffer, rhs.raw_buffer, used_byte_size);
+
+					// set individual pointers to new values
+					size_t base_offset = 0;
+					size_t offset = 0;
+					for (int i = 0; i < batch_cnt; ++i)
+					{
+						shader_prgms = reinterpret_cast<ShaderPrgmData*>(raw_buffer);
```
reinago: no offsets?

plugins/ng_mesh/src/NGMeshRenderer.cpp
```diff
@@ -0,0 +1,256 @@
...
+	for (auto& render_batch : m_render_batches)
+	{
+		render_batch.shader_prgm->Enable();
+
+		render_batch.mesh_shader_params->bind(0);
+		render_batch.mtl_shader_params->bind(1);
+
+		render_batch.draw_commands->bind();
+		render_batch.mesh->bindVertexArray();
+		/*
```
reinago: ?

reinago (changes requested):
- it does not work yet, right? why merge now?
- Readme.md please if you merge
- I don't really like Pascal-style brackets ;)

plugins/ng_mesh/include/ng_mesh/CallNGMeshRenderBatches.h
invor: Fixed in f8bfeb7

plugins/ng_mesh/src/NGMeshRenderer.cpp
invor: What seems to be the problem? glMultiDrawElementsIndirect requires draw commands buffer (SSBO in this case) + geometry to work with given by a vertex array object.

plugins/ng_mesh/include/ng_mesh/CallNGMeshRenderBatches.h
invor: Edit: Fixed in 9863f3b

tobiasrau (changes requested):
- [x] Should be merged with the actual master.
 - [x] Build checks missing
 - [x] glload is deprecated
 - [x] externals should be stored in a external subdirectory in the plugin, not in the plugins `src` directory
 - [ ] `Readme.md` is empty

tobiasrau:
Cleaned and splitted into #314 and #315.

PR #46, opened 2017-09-21
Added the protein_uncertainty-plugin.

schatzkn:


tobiasrau (changes requested):
<s>- [ ] Remove binaries</s>
<s>- [ ] Let CMake download the binaries</s>

plugins/protein_uncertainty/cache/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
tobiasrau, line 2: Is this file mandatory?

tobiasrau (changes requested):
dssp64 is too large for a Repository include

Issue #49, opened 2017-09-27
OSPRay and other renderers


reinago:
OSPRay should write a depth buffer to be correctly combined with other OpenGL output. See, for example, in combination with the CinematicCamera, where manipulators and the OSPRay output fight for depth.

PR #61, opened 2017-10-09
Ospray and OpenGL composition using depth

tobiasrau:


tobiasrau:
Addresses issue #49

tobiasrau:
Does not work with the MuxRenderer

PR #67, opened 2017-10-19
CSV, float table and clang-format

schulzch:


plugins/mmstd_datatools/src/floattable/FloatTableJoin.cpp
```diff
@@ -0,0 +1,159 @@
...
+        // request frame data
+        firstInCall->SetFrameID(outCall->GetFrameID());
+        secondInCall->SetFrameID(outCall->GetFrameID());
+
+        // issue calls
+        if (!(*firstInCall)()) return false;
+        if (!(*secondInCall)()) return false;
+
+        if (this->dataHash != firstInCall->DataHash() || this->frameID != firstInCall->GetFrameID()) {
+            this->dataHash = firstInCall->DataHash();
```
reinago: we need to store both hashes separately, otherwise this blows up. also, all three frameIDs need to be the same

reinago (commented):
I think there is a bug in the hash/frameID check.
Also, on github, the indentation looks weirdly off in many places. How come?

PR #70, opened 2017-10-23
Added Plugin for the Molecular Sombrero Visualization

schatzkn:


schatzkn:
This pull request is really ugly. Therefore it is closed and subseded by a more clean one that adds the modules introduced here to already existing plugins

PR #81, opened 2017-11-16
Core Refactoring

gralkapk:
Moving unnecessary files from core to more suitable plugins.


tobiasrau:
Maybe we should start over again.

PR #107, opened 2018-02-21
Extended Transferfunction Editing

gralkapk:


tobiasrau (changes requested):
Looks good, but the configurator throws an "index out of bounds" error while klicking on the transferfunction editor a second time

PR #115, opened 2018-02-28
Tracking

FlorianFriess:
Updated the old SVN tracking plugin and added it to the "new" MegaMol.


tobiasrau:
Moved to https://github.com/UniStuttgart-VISUS/megamol-closed-plugins

PR #163, opened 2018-07-26
Frame sync in FBO comm

gralkapk:


gralkapk:
@braunms Do we need this with the current compositor/transmitter?

PR #214, opened 2018-10-11
Module for Particle visibility from volume

reinago:
seems to work, but needs testing


plugins/mmstd_datatools/src/ParticlesToDensity.cpp
```diff
@@ -178,20 +187,25 @@ bool datatools::ParticlesToDensity::getDataCallback(megamol::core::Call& c) {
     metadata.MinValues[0] = this->minDens;
     metadata.MaxValues = new double[1];
     metadata.MaxValues[0] = this->maxDens;
-    auto bbox = inMpdc->AccessBoundingBoxes().ObjectSpaceBBox();
+    auto bbox = inMpdc->AccessBoundingBoxes().ObjectSpaceClipBox();
```
gralkapk: Maybe revert to this?

```suggestion
    auto bbox = inMpdc->AccessBoundingBoxes().ObjectSpaceBBox();
```

gralkapk (changes requested):

gralkapk (approved):

PR #222, opened 2018-10-23
tentative fix for up vector mess

reinago:
- seems to fix problems with raycastvolumerenderer as well
- what happens on the powerwall?


gralkapk:
TODO: Test this with Dischingen WASD

gralkapk:
@reinago Not necessary with the new camera, right?

gralkapk:
This PR can be closed, right?

PR #243, opened 2019-01-18
Clean up main() startup

geringsj:
These patches refactor the main() startup code in Console.cpp to be a bit more tidy.


geringsj:
The recent commits clean up and refactor different aspects of the WindowManager, gl::Window and the UI Layers. This simplifies migration to the new frontend infrastructure targeting abstracted/encapsulated GL render contexts.

PR #250, opened 2019-02-01
Graph traversal cleanup

reinago:
Some changes to the way the graph is traversed. Changed FindModuleNoLock, added EnumModulesNoLock

### DIFF
--- a/core/src/CoreInstance.cpp
+++ b/core/src/CoreInstance.cpp
@@ -2009,6 +2009,62 @@ void megamol::core::CoreInstance::LoadProject(const vislib::StringW& filename) {
 }
 
 
+void megamol::core::CoreInstance::EnumModulesNoLock(
+    core::AbstractNamedObject* entry_point, std::function<void(Module*)> cb) {
+    
+    AbstractNamedObject *ano = entry_point;
+    bool fromModule = true;
+    if (!entry_point) {
+        ano = this->namespaceRoot.get();
+        fromModule = false;
+    }
+
+    auto anoc = dynamic_cast<AbstractNamedObjectContainer*>(ano);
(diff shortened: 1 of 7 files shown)


schulzch:
Sieht gut aus... aber willst du wirklich, dass die Enum-Funktion die `Module`-Pointer ohne `const` rausgibt?

core/src/LuaState.cpp
```diff
@@ -1565,6 +1523,43 @@ int megamol::core::LuaState::ListInstatiations(lua_State* L) {
     return 0;
 }
 
+int megamol::core::LuaState::ListParameters(lua_State* L) {
+    if (this->checkRunning(MMC_LUA_MMLISTPARAMETERS)) {
+
+        const int n = lua_gettop(L);
+
+        // TODO I am not sure whether reading information from the MegaMol Graph is safe without locking
```
gralkapk, line 1531: That depends on the execution order wrt. the frame lock.
If script commands are executed asynchronously, a graph lock is necessary to avoid race conditions.
If script commands are executed immediately after graph updates have been performed, a graph lock is not required.

core/src/CoreInstance.cpp
```diff
@@ -2009,6 +2009,66 @@ void megamol::core::CoreInstance::LoadProject(const vislib::StringW& filename) {
 }
 
 
+void megamol::core::CoreInstance::EnumModulesNoLock(
+    core::AbstractNamedObject* entry_point, std::function<void(Module*)> cb) {
```
gralkapk, line 2013: This signature implies that we can provide lambdas that alter a Module.

gralkapk (commented):
Passt scho'

PR #255, opened 2019-02-28
GUI

braunms:
- Improved usability of GUIRenderer (key handling, functionality, fonts ...).
- Replaced use of vislib::sys::KeyCode with new megamol::core::view::KeyCode in Input.h.
- Disabled AntTweakBar via switch in CMake (default is OFF).
- Moved get/set LastFrameTime() from AbstractCallRender3D.h to AbstractCallRender.h to be used in View2D, too.


braunms:
Linux: Successfully compiled and tested using Fedora 29 and gcc versions 5.5 and 8.2 but with disabled InfoVis-Plugin (ScatterplotMatrixRenderer2D.cpp throws type conversion errors).

PR #259, opened 2019-03-07
GUI

braunms:
Some major changes: gui can be used in different view instances independently, drag and drop of module's parameters to new separate window (maybe usable in split views), added user input checks, ...


braunms:
screeny

PR #262, opened 2019-03-15
GUI

braunms:
- Added separate slot for using GUI as SplitView overlay.
- Added linear transfer function parameter (value is JSON string)  and modified LinearTransferFunction module to work with new LTF editor in the gui.


braunms:
transfer_function
Format of transfer Function in JSON: 
{"Interpolation":"LINEAR","Nodes":[[0.0,0.0,0.0,0.006666660308837891,0.0],[0.0,1.0,0.0,1.0,0.09603573381900787],[0.75,0.013333320617675781,0.75,1.0,0.19599999487400055],[1.0,1.0,1.0,1.0,1.0]],"TextureSize":128}

PR #274, opened 2019-03-30
FlowVis

straubar:
So far only computation of implicit vector field topology.


straubar:
- [x] Adjust CUDA to new CMake, including it as a language feature

PR #282, opened 2019-04-15
First version of ballot functor to guard MPI methods

gralkapk:


gralkapk:
Pobably not required due to synced data source mechanics.

tobiasrau:
Implemented different concept.

PR #322, opened 2019-07-12
Moved KHR Debugging (Stack Dump) from Console into Core.

braunms:


core/CMakeLists.txt
```diff
@@ -198,6 +199,23 @@ if(BUILD_CORE)
 
   # Target definition
   add_library(${PROJECT_NAME} SHARED ${public_header_files} ${header_files} ${shader_files} ${source_files} ${version_files})
+
+  # add libunwind
+  if(UNIX)
+    option(USE_LIBUNWIND "use libunwind" ON)
```
gralkapk: Should be default off. Not every machine has libunwind per default and we only need it for debugging.

gralkapk (changes requested):

core/CMakeLists.txt
braunms: Good point. I disabled the option in commit 2e17685.

tobiasrau (approved):

PR #326, opened 2019-07-15
Screenshots are as good as project files now

reinago:
- fix: trailing zero in EXIF
- fix: multi-line parameter values in EXIF
- use EXIF data as lua project for reproducibility goodness


core/src/CoreInstance.cpp
```diff
@@ -1961,6 +1959,48 @@ void megamol::core::CoreInstance::LoadProject(const vislib::StringA& filename) {
...
+                            vislib::sys::Log::DefaultLog.WriteError(vislib::sys::Log::LEVEL_INFO,
+                                "Failed loading project file \"%s\": %s", filename.PeekBuffer(), result.c_str());
+                        } else {
+                            this->loadedLuaProjects.Add(
+                                vislib::Pair<vislib::StringA, vislib::StringA>(filename, content));
+                        }
+                    } else {
+                        vislib::sys::Log::DefaultLog.WriteError(
+                            "LoadProject: Unable to extract png exif data");
+                    }
```
braunms: The **png_infop** could be cleaned up here with:
**png_destroy_info_struct (png, &info);**

core/src/CoreInstance.cpp
```diff
@@ -1961,6 +1959,48 @@ void megamol::core::CoreInstance::LoadProject(const vislib::StringA& filename) {
...
+                }
+                fclose(fp);
+                png_destroy_read_struct(&png, &info, nullptr);
+                // TODO: dispose exif_data buffer...?
```
braunms: What are the arguments against **delete [] exif_data;** ?

core/src/CoreInstance.cpp
reinago: that code should already be there...

core/src/CoreInstance.cpp
reinago: I did not have time to test that. I have no idea whether delete or delete[] is appropriate here.

core/src/CoreInstance.cpp
```diff
@@ -1961,6 +1959,48 @@ void megamol::core::CoreInstance::LoadProject(const vislib::StringA& filename) {
...
+                    if (exif_size > 0) {
+                        const vislib::StringA content(reinterpret_cast<char*>(exif_data));
+                        std::string result;
+                        //vislib::sys::Log::DefaultLog.WriteInfo(
+                        //    "Loaded project from png:\n%s", content.PeekBuffer());
+                        if (!this->lua->RunString(content.PeekBuffer(), result, filename.PeekBuffer())) {
+                            vislib::sys::Log::DefaultLog.WriteError(vislib::sys::Log::LEVEL_INFO,
+                                "Failed loading project file \"%s\": %s", filename.PeekBuffer(), result.c_str());
+                        } else {
+                            this->loadedLuaProjects.Add(
```
gralkapk: Why VISlib :(

core/src/CoreInstance.cpp
```diff
@@ -1997,6 +2036,46 @@ void megamol::core::CoreInstance::LoadProject(const vislib::StringW& filename) {
                     vislib::Pair<vislib::StringA, vislib::StringA>(vislib::StringA(filename), content));
             }
         }
+    } else if (filename.EndsWith(L".png")) {
+        png_structp png = png_create_read_struct(PNG_LIBPNG_VER_STRING, NULL, NULL, NULL);
+        if (!png) {
+            vislib::sys::Log::DefaultLog.WriteError("LoadProject: Unable to create png struct");
+        } else {
+            FILE* fp = fopen(W2A(filename.PeekBuffer()), "rb");
```
gralkapk: Isn't that the only difference to the above version?

gralkapk (changes requested):

core/src/view/special/ScreenShooter.cpp
```diff
@@ -494,7 +494,7 @@ void view::special::ScreenShooter::BeforeRender(view::AbstractView* view) {
                                 val.replace(start_pos, from.length(), to);
                                 start_pos += to.length(); // Handles case where 'to' is a substring of 'from'
                             }
-                            confParams << "mmSetParamValue(\"" << slot->FullName() << "\",\"" << val << "\")\n";
+                            confParams << "mmSetParamValue(\"" << slot->FullName() << "\",[[" << val << "]])\n";
```
braunms: Unescaped quotes don't need and even shouldn't be escaped any more when multiline string delimiter ist used (Escaped quotes break JSON string, e.g.). And delimiter shouldn't be "[[" since this can be part of a JSNON string, too (use e.g. "[=[" and "]=]"?).

tobiasrau (approved):

PR #351, opened 2019-07-23
revamped flagstorage/call

reinago:
- moved to core so it can be used by all modules
- new unique-pointer-like paradigm: the flag data is only in one place at once and you have to give it back to the call/source when you are done
- preliminary flag support in ArrowRenderer, needs to be fixed as soon as selection works again
- updated ParallelCoordinateRenderer


reinago (commented):
good enough

tobiasrau (approved):

PR #356, opened 2019-07-26
mmpldinfo can compute local bounding boxes

reinago:


tobiasrau (changes requested):
Some plugins are missing the CMakeLists.txt changes.

tobiasrau (approved):

PR #363, opened 2019-07-30
GUI: Revert string cropping fix.

braunms:


braunms:
Will be fixed in #366 ...

Issue #333, opened 2019-07-17
GUIView is not a complete implementation of AbstractView


reinago:
- example: crash when you try to make a screenshot of a GUIView: it does not implement OnRender.
TODO: find out what else is missing

reinago:
- [x] onRenderView #367
- [x] isParamRelevant
- [x] DesiredWindowPosition
- [x] unpackMouseCoordinates

braunms:
isParamRelevant() and DesiredWindowPosition() are already implemented in AbstractView.
unpackMouseCoordinates is now implemented in PR #383

straubar:
@reinago Can you cherry-pick the commits from the astro branch? Or did Matthias already fix everything?

braunms:
The changes concerning onRenderView we discussed should already have been merged in #366.

PR #367, opened 2019-08-02
Mostly bug fixes

straubar:


straubar:
Does this solve #333 @reinago ?

reinago:
It should fix the ScreenShooter issue, but, as far as I can see, it is still incomplete:
- isParamRelevant
- DesiredWindowPosition
- unpackMouseCoordinates
are still missing. Also, this needs to go into master-dev ASAP

PR #376, opened 2019-08-07
GUI Multi Line String Input Fix

braunms:


plugins/gui/src/GUIView.cpp
```diff
@@ -1629,19 +1630,16 @@ void GUIView::drawParameter(const core::Module& mod, core::param::ParamSlot& slo
             // Determine multi line count of string
             int lcnt = static_cast<int>(std::count(it->second.begin(), it->second.end(), '\n'));
             lcnt = std::min(static_cast<int>(GUI_MAX_MULITLINE), lcnt);
```
reinago, line 1632: looks good, but I still wonder why you do not seem to need a max(1,lcnt) as well? GUI elements with no height do not seem to happen, but I don't understand why

tobiasrau (approved):

plugins/gui/src/GUIView.cpp
braunms: ImGui interprets values of zero as "auto size".

PR #386, opened 2019-08-13
SphereRenderer

braunms:
- Added FlagStorage support for all render modes
- Consolidated redundant shader code
- Added Ambient Occlusion to regression test
- Some code formatting and cleanup
- Table2Particles: Considering radius for bbox size calculation
- Transformed color param in ClippingPlane to 'real' ColorParam


reinago:
can we consolidate this even more? I see many blah_attributes and they are pretty similar. also sometimes the blah start still contains attributes, doesn't it?

core/Shaders/sphere/splat_fragment_mainstart.glsl
```diff
@@ -0,0 +1,3 @@
+
+void main(void) {
```
tobiasrau: Could this also become a simple_mainstart?

tobiasrau (changes requested):

core/Shaders/sphere/splat_fragment_mainstart.glsl
braunms: Jupp, good point, done in commit a31ee39.

tobiasrau (approved):

tobiasrau (changes requested):
1|[SphereRenderer] Unable to compile sphere shader (@compile vertex): 0:16(2): error: vertex shader input / attribute cannot have array type in GLSL 1.30 (GLSL 1.50 required)

PR #437, opened 2019-08-30
improve build speed

reinago:
Build process is so boring.
- Tried the UPDATE_COMMAND cheat.
- TODO: put FetchContent in place


straubar:
Switched to FetchContent and centralizing all external dependencies.
Now, external libraries are downloaded and configured while running cmake. Additionally, the update is disconnected and therefore already downloaded versions are never updated.

schulzch:
The `_ext` suffix was only meant to work around name collisions between `add_library()` and `ExternalProject_Add()`, i.e., target references in `target_link_libraries(...)` should never have any `_ext` suffixes.

straubar:
> 
> 
> The `_ext` suffix was only meant to work around name collisions between `add_library()` and `ExternalProject_Add()`, i.e., target references in `target_link_libraries(...)` should never have any `_ext` suffixes.

The new method creates an alias target, which can be used after calling to require_external. This way, the decision to use a static or shared library will be made globally and the user can just use `<REQUESTED_LIB>_ext` as target. Without the `_ext` suffix, it can happen that the alias has the same name as another target. E.g., zlib creates a target `zlib` and `zlibstatic`. Using the static version with alias `zlib` might lead to undefined behavior.

straubar:
See #445 for new, better version

PR #444, opened 2019-09-06
CMake cleanup all over the place

schulzch:
This PR catches a lot of wild growth regarding CMake:
- let CMake 2.x properly error (not just a warning)
- move stuff from root `CMakeLists.txt` to `megamol_options.cmake`
- allow clang-format 8.x and 9.y (tested)
- globally enable C++17 (we use it anyway all over the place)
- rework crusty MPI stuff
- rework weird OpenMP stuff (spot tested g++ and msvc)
- added a test around threading for clang (that one is very weird)
- remove add_dependencies(), include_directories(), and add_definitions() all over the place
- drop Eigen from molecularmaps
- migrate all plugins to the same string(TOUPPER ...) option(...) pattern
- drop/replace some superfluous/meaningless comments
- drop commented blocks of CMake code (we have a version control system and enough example code to steal from)
- alignment all plugin-related CMakeLists.txt files (execpt for source grouping)

The next step would be to align source grouping and enable clang-format all over the place.


tobiasrau (dismissed):
Should be OK after resolving the merge conflicts.

tobiasrau (approved):

PR #445, opened 2019-09-06
New handling of external libraries

straubar:
Similar to #437

Ninja support:
- [x] Tested

Only tested with the default turned on plugins. The following need to be adapted:
- [x] zfp
- [x] tracking (private repository and therefore not tested)
- [x] adios2
- [x] snappy
- [x] IceT

Platforms:
- [x] Tested on Linux
- [x] Tested with all plugins on Linux
- [x] Tested on Travis
- [x] Tested on Appveyor

Todo:
- [x] Install shared libraries into the MegaMol install path

Only downside I can see so far is that initial configure step takes a lot of time now. Subsequent ones not, however.


schulzch:
Ninja has to know how to obtain a dependency, i.e., the BYPRODUCT has to be specified. I guess you want to add it to the `add_custom_target()` call.

schulzch:
> Only downside I can see so far is that initial configure step takes a lot of time now. Subsequent ones not, however.

Isn't this exactly what we want (move build time to configuration time). If it only slows down the frist configuration, but not subsequent ones: even better! Anyway, do we have (handweavy) numbers on this?

tobiasrau (approved):

PR #450, opened 2019-09-09
Fixed options

straubar:
Fixes gcc build error


straubar:
Appveyor pulled the commit from the wrong pull request. There should be no reason for it not to work on Windows using Visual Studio.

tobiasrau (approved):

PR #456, opened 2019-09-23
Camera manipulator update pt.2

invor:
Tried to unify and simplify manipulator handling and usage. In theory only one mouse controlled manipulator should be active at any given time now.

Added panning to orbital manipulators (arcball and turntable) with middle mouse button hold + mouse drag.

MoveStep and AngleStep of View3D_2 are now given in scene units per second and degree per second respectively.

More cleaning up, further simplifications and additinal features hopefully will follow in a pt.3. The "stuttering camera" bug remains yet to be identified and fixed...


invor:
Added some comments to changelog

schulzch:
Please do not comment code, but delete it. We have a commit history to restore it.

reinago:
- when the arcball is set as default, ctrl does not work
- turntable should probably fix the y axis when you start manipulating
- mouse wheel for dolly?

invor:
> 
> 
> Please do not comment code, but delete it. We have a commit history to restore it.

Done

invor:
> 
> 
>     * when the arcball is set as default, ctrl does not work
> 
>     * turntable should probably fix the y axis when you start manipulating
> 
>     * mouse wheel for dolly?

Arcball default + CTRL fixed and mouse dolly for orbital cameras fixed in 4d34818ef470a1292ff26da986128dc581669fb8

Thinking about turntable up-axis again: As it is primarily intended to be used for datasets with a meanigful up-direction, I believe it should remain fixed around the world-up axis and not updated after camera movements. Mixing arcball/Freelancer and turntable should propably be simply discouraged...

tobiasrau (approved):

PR #457, opened 2019-09-23
External dependencies are built as shared objects (so/dll)

### LINKED ISSUE #455: #445 merge killed ADIOS2
The ADIOS implementation started to throw bad_allocation exceptions.
Reverting to a state before the build_speed PR #445 made ADIOS to work again.
Interestingly, using an external built ADIOS2 causes the same issues.

straubar:
This should resolve #455


straubar:
As all externals are built as shared objects and linked dynamically, it is possible to reduce them to Release builds. Thus, configuration and compilation times are reduced. Additionally, the install mechanism for external libraries was centralized, as all need to be installed into the appropriate binary directory. Sources and symbols are excluded from installation.

schulzch:
This is a **very bad idea**. Dynamic libraries will hurt run-time performance. Really, really bad idea as in I expect performance sadness -30%

straubar:
Can someone test these changes to the build system? I can't imagine a significant loss in performance, though.

PR #463, opened 2019-10-30
Internal refactoring of Cinematic Plugin

braunms:
- Switched keyframe format to JSON (in changelog)
- Adapted to new camera
- Replaced legacy OpenGL and switched to glm
- Added CinematicUtils to pool rendering stuff and utility functions
- Fixed SDFFont rotation for batch rendering
- Fixed input propagation for RenderModule
- Added project export into header of all rendered png images


core/include/mmcore/LuaInterpreter.h
```diff
@@ -203,6 +203,7 @@ std::string const megamol::core::LuaInterpreter<T>::DEFAULT_ENV =
     "  next = next,"
     "  pairs = pairs,"
     "  pcall = pcall,"
+    "  require = require,"
```
reinago: I'm not sure I remember correctly, but didn't we want to work around `require` by using the internal file loading via `mmReadTextFile`? that way we at least _could_ sandbox that if we ever need to

plugins/cinematic/README.md
```diff
@@ -60,9 +60,9 @@ The module `KeyframeKeeper` exposes the following parameters:
 * `editSelected - animTime` (`1.0`): Edit animation time of the selected keyframe.
 * `editSelected - simTime` (`1.0`): Edit simulation time of the selected keyframe.
 * `editSelected - positionVector`: Edit the position vector of the selected keyframe.
-* `editSelected - lookatVector`: Edit the look-at vector of the selected keyframe.
-* `editSelected - resetLookat` (Assigned key: `CTRL+l`): Reset the LookAt vector of the selected keyframe.
-* `editSelected - upVector`:  Edit the up vector of the selected keyframe.
+* `editSelected - viewVector`: Edit the view vector of the selected keyframe.
```
reinago: shouldn't we just use the notation you can see in MegaMol anyway? Like `editSelected::lookatVector` ? Or how does this appear in the GUI...

tobiasrau (approved):

Issue #480, opened 2019-11-21
Configurator cannot parse strings containing "\" when opening LUA projects


straubar:
When opening a LUA project, which sets a transfer function string containing escaped characters like \" for quotation marks, the reading of the string ends after the backslash.

As an example see in the megamol-examples repository: volume/raycast_volume_renderer.lua

PR #481, opened 2019-11-26
fixed an issue where parameter values that include double quotes are …

sellendk:
…not properly imported. This fixes the issue from Issue #480


straubar (approved):
This successfully fixes the problem.

PR #483, opened 2019-11-26
Bugfixes and features

### LINKED ISSUE #474: SetSlotUnavailable triggers Assert
Calling SetSlotUnavailable (always?) triggers an assert when checking whether the parent module of the slot matches the module it is about to be remove from:
`ASSERT(child->Parent().get() == this);`

This is most likely caused by modules not being fully created yet when slots are being added (i.e. made available) to modules in the constructor. Either this has to be fixed, or the assert needs to be replaced with a more appropriate safety check.

Maybe this issue will be automatically solved by the new graph?

### LINKED ISSUE #477: BoundingBoxRenderer does not set camera in first frame
This is leading to an exception, which is silently ignored.

### LINKED ISSUE #484: VolumetricDataSource kills performance by computing dataset min/max
Performance of the RaycastVolumeRenderer suffers considerably because the VolumetricDataSource seems to recalculate the min/max value of the volume in each onGetData() call.

A while back a simple dataset like bonsai could be rendered with performance suitable for stereo VR exploration of the dataset. Currently bonsai is rendered with 21 fps (mono).

megamol (Debugging) - Microsoft Visual Studio  26 11 2019 19_18_42

straubar:
Fixed the following bugs:
- Wrong include guard in DataHash
- OpenGL state was left in bad shape in SplitView (how did this ever work correctly?)
- Camera propagation of the BoundingBoxRenderer is also done in the Render method, which fixes #477 
- VolumetricDataSource was reloading the input file every frame, this is now prevented by better hash usage, which should fix #484 

Added feature:
- BoundingBoxRenderer can now render a (basic) view cube, with user-defined position (bottom left, bottom right, top left, top right) and size

Additionally, I removed the assert for SetSlotUnavailable, as it does not seem to have any positive effect, disallowing removal of slots in the constructor (closes #474).

### DIFF
--- a/core/src/view/BoundingBoxRenderer.cpp
+++ b/core/src/view/BoundingBoxRenderer.cpp
@@ -197,6 +197,9 @@ bool BoundingBoxRenderer::Render(CallRender3D_2& call) {
             "The BoundingBoxRenderer does not work without a renderer attached to its right");
         return false;
     }
+    *chainedCall = call;
+    bool retVal = (*chainedCall)(view::AbstractCallRender::FnGetExtents);
+    call = *chainedCall;
 
     Camera_2 cam;
     call.GetCamera(cam);
--- a/core/src/view/BoundingBoxRenderer.cpp
+++ b/core/src/view/BoundingBoxRenderer.cpp
@@ -336,26 +336,7 @@ bool BoundingBoxRenderer::RenderViewCube(CallRender3D_2& call) {
     call.GetCamera(cam);
 
     const auto orientation = cam.orientation();
-
-    // Calculate rotation matrix from quaternion,
-    // see https://en.wikipedia.org/wiki/Quaternions_and_spatial_rotation#Quaternion-derived_rotation_matrix
-    const auto qr = orientation.w();
-    const auto qi = orientation.x();
-    const auto qj = orientation.y();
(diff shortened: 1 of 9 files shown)


geringsj (dismissed):

core/src/view/BoundingBoxRenderer.cpp
```diff
@@ -288,4 +327,96 @@ bool BoundingBoxRenderer::RenderBoundingBoxBack(const glm::mat4& mvp, const Boun
...
+    const auto orientation = cam.orientation();
+
+    // Calculate rotation matrix from quaternion,
+    // see https://en.wikipedia.org/wiki/Quaternions_and_spatial_rotation#Quaternion-derived_rotation_matrix
+    const auto qr = orientation.w();
+    const auto qi = orientation.x();
+    const auto qj = orientation.y();
+    const auto qk = orientation.z();
+
+    glm::mat4 rotation(1.0f);
```
geringsj: To calculate a rotation matrix from a quaternion one could use glm::quat_cast here, see https://glm.g-truc.net/0.9.0/api/a00135.html#ae2d14d52f9fef3da00cf867f9e2e3dfa

geringsj (dismissed):

geringsj (approved):

PR #486, opened 2019-12-04
Examples now cloned to install path

tobiasrau:


straubar (changes requested):
I would suggest cloning the examples repository into the CMAKE_BINARY_DIR and adding the directory to the CMake install mechanism. This way, changing the CMAKE_INSTALL_PREFIX variable after the initial configuration step, does not lead to downloading the examples again.

straubar:
I would suggest installing into share/examples. However, this means that all paths in the example projects have to be adjusted.
Please state an opinion on what is the preferred way: installing into examples or share/examples.

Also, issue #489 is somehow related to the examples and a possible need for refactoring.

straubar (dismissed):

tobiasrau:
I'm not sure why to hide the examples in the share directory and not simply put them to `CMAKE_INSTALL_PREFIX`. The examples should be self-consistent and not dependent on some external paths.

moritz-h:
@tobiasrau The current path names schema comes obviously from Linux world. When you want to adapt this schema it is very bad style to put thinks directly in `CMAKE_INSTALL_PREFIX`. Correct place for this would be something like share/megamol/examples
(CMAKE_INSTALL_PREFIX usually defaults to /usr/local, so you normally want the absolute path to files to be `/usr/local/share/megamol/stuff` and not put the examples to `/usr/local/examples`)

But I don't know how this will complicate stuff for Windows world. So if there are good other reasons for a different schema you can change it of course. But I would suggest to change the `CMAKE_INSTALL_PREFIX` default to something like `/opt/megamol/` in that case. Then you use whatever structure you like.

tobiasrau:
The `CMAKE_INSTALL_PREFIX` of MegaMol does not default to some system path.

straubar (approved):

PR #487, opened 2019-12-04
External dependencies are built only once

### LINKED ISSUE #455: #445 merge killed ADIOS2

straubar:
External dependencies are downloaded and configured only once during the CMake configuration step. Reconfiguration can be triggered manually by setting a specific CMake variable, and is triggered automatically if the git repository or tag has changed.

Building is performed during the building process. Shared libraries are always built as Release, static libraries are built according to the settings (configuration, multi-configuration system). Once an external dependencies has been built, it will not be rebuilt unless reconfiguration is triggered as described above. Static libraries are only "installed" within the build directory for preventing a clash of different configurations. Shared libraries are "installed" first within the build directory.

In the installation step, shared libraries are copied from their build install directory into the output MegaMol directory.

(This should resolve #455, supercede #457, and include #476)


CMakeExternals.cmake
```diff
@@ -135,52 +140,54 @@ function(require_external NAME)
...
 
-    if (MSVC)
-      set(GLFW_PRODUCT "lib/glfw3dll.lib")
-      set(GLFW_IMPORT_LIBRARY "lib/glfw3dll.lib")
-      set(GLFW_LIBRARY "lib/glfw3.dll")
+    if(WIN32)
+      set(GLFW_IMPORT_LIB "lib/glfw3dll.lib")
+      set(GLFW_LIB "lib/glfw3.dll")
+      set(MOVE_CMD COMMANDS COMMAND ${CMAKE_COMMAND} -E copy "\"<INSTALL_DIR>/lib/glfw3.dll\" \"<INSTALL_DIR>/bin/glfw3.dll\""
+                            COMMAND ${CMAKE_COMMAND} -E remove -f \"<INSTALL_DIR>/lib/glfw3.dll\")
```
tobiasrau, line 170: This is a bit too specific.

CMakeExternals.cmake
```diff
@@ -463,67 +443,23 @@ function(require_external NAME)
...
-    add_external_library(natnet SHARED 
-      PROJECT tracking
-      IMPORT_LIBRARY_DEBUG ${TRACKING_NATNET_IMPORT_LIB}
-      IMPORT_LIBRARY_RELEASE ${TRACKING_NATNET_IMPORT_LIB}
-      LIBRARY_DEBUG ${TRACKING_NATNET_LIB}
-      LIBRARY_RELEASE ${TRACKING_NATNET_LIB})
-
-    add_external_library(tracking_int INTERFACE
-      PROJECT tracking
-      INCLUDE_DIR "src/tracking_ext/tracking/include")
```
tobiasrau: Why is the tracking gone?

tobiasrau (changes requested):
The ADIOS plugin still does not work as intended.

CMakeExternals.cmake
straubar: The glfw dll is installed into the lib directory instead of the dll directory. Thus, not copying it to the correct folder, the MegaMol executables are not able to find it. Therefore, this has to be done at some point.

straubar:
> 
> 
> The ADIOS plugin still does not work as intended.

This problem only occurs if MegaMol is built in Debug configuration. The only feasible work-around I can see is to build adios2 as a static library.

CMakeExternals.cmake
straubar: Because it is a private repository and trying to clone it leaves CMake hanging.

straubar: Now that mm-tracking is public, I readded tracking to the CMakeExternals.

tobiasrau (approved):

Issue #478, opened 2019-11-19
Configurator Exception with lua string


moritz-h:
Configurator cannot open projects if TableManipulator lua sources contains a `<` char, like for example the default source of the TableManipulator.
Editing and saving is no problem, but then the project cannot be loaded again.

```
System.Xml.XmlException: '<', hexidezimaler Wert 0x3C, ist ein ungültiges Attributszeichen. Zeile 2114, Position 14.
   bei System.Xml.XmlTextReaderImpl.Throw(Exception e)
   bei System.Xml.XmlTextReaderImpl.ParseAttributeValueSlow(Int32 curPos, Char quoteChar, NodeData attr)
   bei System.Xml.XmlTextReaderImpl.ParseAttributes()
   bei System.Xml.XmlTextReaderImpl.ParseElement()
   bei System.Xml.XmlTextReaderImpl.ParseElementContent()
   bei System.Xml.XmlLoader.LoadNode(Boolean skipOverWhitespace)
   bei System.Xml.XmlLoader.LoadDocSequence(XmlDocument parentDoc)
   bei System.Xml.XmlDocument.Load(XmlReader reader)
   bei System.Xml.XmlDocument.Load(String filename)
   bei MegaMolConf.Io.ProjectFile.Load(String filename) in C:\Projects\MegaMol\megamol\configurator\MegaMolConf\Io\ProjectFile.cs:Zeile 57.
   bei MegaMolConf.Program.TryLoadRunFile(String path) in C:\Projects\MegaMol\megamol\configurator\MegaMolConf\Program.cs:Zeile 178.
```

moritz-h:
Bug is not only in configurator, also mmconsole.exe is affected and cannot parse the xml

straubar:
There seem to be multiple problems:
- mmprj-files are written without masking characters, such as `<` as `&lt;`. By doing so manually, the project files can be loaded in the configurator and in MegaMol
- mmprj-files with the masked characters can be loaded, however, fail to produce the correct result (without error, but in my case no values arrived at the PCR on the left side of the table manipulator)
- lua-files containing the line `mmSetParamValue("::Project_1::TableManipulator1::script", "")` cannot be loaded in the configurator even with empty initialization string (weird IndexOutOfRangeException)
[...]

Issue #489, opened 2019-12-06
Paths in LUA project files


straubar:
As in our examples repository, LUA files sometimes store (relative) paths using `/` as separator and sometimes using `\\`. On Windows, having paths with `/` separators as `FilePathParam`, it does not work. Probably the contrary is true for Linux systems.

straubar:
This bug seems to have vanished (or never existed).

PR #507, opened 2020-01-21
Configurator

sellendk:
Fixes problems from Issue #478 and #489.
Also added a new button to the configurator on the top right. Clicking that button hides all call-names and its rectangles.


straubar (approved):
Seems to fix #478, the minimal example provided works with the proposed changes.
Hiding the call names works as described.

PR #530, opened 2020-03-10
Configurator [PROTOTYPE]

braunms:
Added configurator prototype window to GUI (see readme for starting configurator as stand-alone project). Plus some general refactoring, cleanup and bug fixing ...


invor:
Just testing (no proper reviewing yet) I found one issue (for my stuff at least) with the following workflow:
1) Start MegaMol with configurater.lua project
2) Load CompositingExample.lua in configurator
3) Layout graph, then save otherwise unchanged graph to a new project file
4) Exit MegaMol, start it again with the newly saved project -> black screen

Note that the original CompositingExample.lua works just fine with the same MegaMol version. Also, starting MegaMol with the CompositingExample, loading the running project in the configurator and then saving it to a new file somehow seems to produce a working project file...

invor:
Some more comments after additional hands-on testing with current version:
- If parts of the graph are moved out of the lower screen boundary, they won't be rendered properly after zooming out until the view is panned
- ~~Happened to me once that when loading a running project, I had two gui views in the project which I noticed when I couldn't save the project (rightly so). However, I couldn't reproduce this bug a second time yet.~~
- Not a bug: Zooming in/out with the scroll wheel is a little slow right now.
- Another feature thing: I guess a more feature-rich file browser would require quite a lot of time? Therefore no priority for current PR, but a "move up a folder" button was greatly missed during testing.

invor (approved):
As expected, the large amount of files and changes makes it nearly impossible to check the code in detail right now. But the more critical bugs found so far by testing have been fixed and I can't find any new ones right now. I say, let's merge it so the prototype can be tested by a larger group :-)

PR #533, opened 2020-03-16
Improve SPLOM picking

moritz-h:
- Add pick radius parameter
- Draw selection marker
- Selection reacts now on first click, not only on move
- Remember selection, drawing with left clicks adds now to selection instead of replacing selection. Therefore it is possible to draw a larger selected area instead of just selecting pick radius around the mouse. Drawing with right clicks, removes selection
- Reset button to clear selection


schulzch (approved):
Looks good.

PR #535, opened 2020-03-17
operator=() for CallGeneric and GenericVersionedCall

moritz-h:


invor (commented):
On first glance the actual code seems fine, however:
1) Could this impact any internal workings of the graph system because it is now longer possible to do a deep copy of a call? I.e. are there any cases where you would need to copy the caller and callee pointers?

2) For the versioned call: This will implicitly synchronise the versioning of calls from left to right. I think this will kill the hasUpdate() function if a higher get_version is assgined coming from left, than the set_version that is assigned afterwards by the actual callback of the call. This will happen as soon as a module on the left side of a connection is updating at a higher frequency than the one on the right. To avoid accidents, I think at the very least, the versions musn't be overwritten. On the other hand, since there is an update to the data, the version should be incremented...

Overall, the more I think about this, the more I would suggest to not copy calls at all like this. It even makes me want to delete some of the constructors and assignment operations for call classes altogether! There is just too much potential for unforeseen consequences. Instead, consider using the built-in setData and setMetaData functions. It's really just one addtional line of code compared to calling the assignment operator and it communicates much more clearly the intent of the action: you don't want a copy of the Call object, you want to pass along the contained data and metadata packages. Also it will keep the versioning as intact as possible.

But I'm open for additional comments and suggestions.

PR #540, opened 2020-03-25
Ospray update

tobiasrau:


plugins/mesh/src/gltf/glTFFileLoader.cpp
```diff
@@ -49,25 +48,19 @@ bool megamol::mesh::GlTFFileLoader::getGltfDataCallback(core::Call& caller) {
...
-    checkAndLoadGltfModel();
+    auto has_update = checkAndLoadGltfModel();
+    if (has_update) {
+        ++m_version;
+    }
 
-    gltf_call->setData(m_gltf_model);
+    if (gltf_call->version() < m_version) {
+        gltf_call->setData(m_gltf_model,m_version);
+    }
```
reinago, line 58: I thought the idea was to wrap this if in the setData?

reinago (commented):
let's first make the comment appear...

plugins/mesh/include/mesh/MeshCalls.h
```diff
@@ -73,22 +74,33 @@ class MESH_API CallGPURenderTaskData
...
-    CallMesh() : CallGeneric<std::shared_ptr<MeshDataAccessCollection>, core::Spatial3DMetaData>() {}
+    CallMesh() : core::GenericVersionedCall<std::shared_ptr<MeshDataAccessCollection>, core::Spatial3DMetaData>() {}
     ~CallMesh(){};
 
     static const char* ClassName(void) { return "CallMesh"; }
     static const char* Description(void) { return "Call that gives access to CPU-side mesh data."; }
 };
 
+class MESH_API CallImage
+    : public core::CallGeneric<std::shared_ptr<ImageDataAccessCollection>, core::BasicMetaData> {
```
invor, line 87: Is CallImage not using the versioned call for a reason?

plugins/mesh/src/gltf/glTFFileLoader.cpp
invor: Also found this pattern in the WavefrontObjLoader. I don't think the if cond. is needed. I also don't think it is needed withing setData, as long as pointer sized values are stored in calls.

I think it might be a leftover pattern from back when setData was simply incrementing the version internally and the version wasn't explicitly set.

invor (changes requested):
No point in me checking the Ospray code... The quirks in handling the versioning within the mesh loader modules seem mostly harmless and can be fixed later on.

However, just ran a quick test with examples/CompositingExample.lua and it doesn't work: Because CallMesh and CallGltf are now versioned, every module that uses them needs to be updated as well. So, we have to cherry-pick GPUMeshes and gltfRenderTasksDataSource from the probe_vis branch...

tobiasrau:
We just pull request the whole branch.

PR #545, opened 2020-04-03
SPLOM: draw cell labels on mouse hover

moritz-h:
Draw labels directly on currently hovered cell. This is very useful when zooming into SPLOM on data with many columns. There is a param for switching this on/off. Default is off.
Example:
splom-mouse-labels


reinago (commented):
I think that you should disable batching and make two "immediate mode draws" for these dynamic strings. I checked with Matthias, the batch should be re-usable across frames (no idea whether SPLOM does that). Matthias also advised that you re-use the font that is already there since it's configured the same anyway... and going to 'immediate mode' as above should work fine.

reinago (approved):
that feature was sorely missing...

PR #546, opened 2020-04-07
SPLOM: update screen on data change

moritz-h:


plugins/infovis/src/ScatterplotMatrixRenderer2D.cpp
```diff
@@ -437,6 +437,8 @@ bool ScatterplotMatrixRenderer2D::validate(core::view::CallRender2D& call, bool
     const size_t colCount = this->floatTable->GetColumnsCount();
 
     if (this->dataHash != this->floatTable->DataHash()) {
+        this->screenValid = false;
```
schulzch: This does not invalidate the screen buffer on `dataTime` changes. 
Maybe, next to ` this->textValid = false;` (L457) would be the right place?

schulzch (approved):

Issue #539, opened 2020-03-24
GUI Plugin: Bug and Feature Tracker


braunms:
## Bug
- [ ] Bug - Add alternative hotkey for scrolling (Laptops have no 'middle mouse' button).
- [ ] Bug - Fix varying layout distance between groups for different zoom level.
- [ ] Bug - ImGui: Flickering when vertical and horizontal scrollbars apear at the same time (in a certain way ...)
- [ ] Bug - If mmSetGUIVisible(fals) is called in docked(!) GUI LogConsole window -> mouse interaction is blocked

## Misc

- [ ] Misc - Allow to overwrite existing screenshot files.
- [ ] Misc - Make embedded lua project in screenshot opt-in.
- [ ] Misc - Allow global GUI widgets to be placed in sub-viewport of corresponding view module (e.g. view cube).

## Feature

- [ ] Feature - Sub-groups (groups can contain other groups recursively).
- [ ] Feature - Group multi-selection and combined selection of modules and groups.
- [ ] Feature - Show configurator fullscreen and use global menu bar (mainly to prevent same menu entries twice).
- [ ] Feature - Optimize Performance? (e.g.: directly sync core parameters via core-pointer)
- [ ] Feature - Dialog for screenshots on exit: do you want to provide a more descriptive prefix for these images?
- [ ] Feature - Provenance: Diffs-per-frame-callback. Separate window with parameter history. +back, forward, branching ...
- [ ] Feature - Lua function editor for modifying parameter values.
- [ ] Feature - Keyframe editor / Move cinematic editor to frontend service
- [ ] Feature - Add possibility to load and sync multiple "running" graphs -> project handling
- [ ] Feature - Render output of graph entires in GUI child window.
- [ ] Feature - Add option to return always relative paths for requested files (mainly for file path params).
- [ ] Feature - Save Profiling Bar state to GUI state
- [ ] Feature - Support mltiple compatible calls between modules -> multi-call pop-up
- [ ] Feature - Separate saving of selected group (via context menu)
- [ ] Feature - ClipPlane Manipulation (rename to Plane) (https://github.com/CedricGuillemet/ImGuizmo) - Move on Axes, Rotate, Preview Plane

---

### Pull Request #1111

- [x] Various fixes due to version upgrade of ImGui (see description of PR for more details)

### Pull Request #1005
[...]

PR #549, opened 2020-04-22
Configurator

braunms:
Added some more features and fixed bugs. See #539
Improved usability.


invor:
Played around with current state (tested in CompositingExample.lua). Tow minor problems:
- File dialog of GlTFFileLoader will not go outside MegaMol root directory with folder up command
- deleting a module inside a group will create a new interface slot for that call. Is that on purpose?

braunms:
Fixed minor problems from above in d2532de:
- Relative path ".." blocks going to parent directory. Only ".." is now converted to absolute path, keeping possibility of using relative paths in file path parameters.
- Adding new interface slots on module deletion was not on purpose :-P

invor (approved):
Tested loading and editing different projects. Everything seems fine so far.

PR #550, opened 2020-04-26
Splom UI extensions

moritz-h:
This PR adds several features for the SPLOM UI
- Option to invert Y axis, so diagonal goes from top left to bottom right instead of bottom left to top right
- Make labels at diagonal optional
- Option to add labels + ticks at outside of the SPLOM Matrix
- Mouse hove labels: move label to left, add tick values
- Set minimum tickLength to 0 instead 0.5
- Fix label placement for different sizes and offsets
- Fix breaking of View2D bounding box by not unbinding texture

TODOs:
=> Moved to #595

splom_ui


schulzch:
Looks good so far! 👍 
- Are there issue regarding label placement in scientific mode?
- The "mouse hover labels" show axis-labels near a plot on mouse hover?
- Text margin/padding looks a bit off (X and Y axis are different)

moritz-h:
>     * Are there issue regarding label placement in scientific mode?
No issues so far, just the new options not implemented yet.

>     * The "mouse hover labels" show axis-labels near a plot on mouse hover?
Yes, but that feature was already there before, see #545 for details. Here only layout is updated.
 
>     * Text margin/padding looks a bit off (X and Y axis are different)
Which margins do you exactly mean? Perhaps lets discuss that one directly.

schulzch (commented):
Looks good

schulzch (approved):

PR #551, opened 2020-04-26
Fix transferfunction + new colormaps

moritz-h:


plugins/gui/src/TransferFunctionEditor.cpp
```diff
@@ -144,7 +151,16 @@ std::array<std::tuple<std::string, PresetGenerator>, 12> PRESETS = {
     std::make_tuple("Cubehelix (default)", CubeHelixAdapter(0.5, -1.5, 1.0, 1.0)),
     std::make_tuple("Cubehelix (default, colorful)", CubeHelixAdapter(0.5, -1.5, 1.5, 1.0)),
     std::make_tuple("Cubehelix (default, de-pinked)", CubeHelixAdapter(0.5, -1.0, 1.0, 1.0)),
-    std::make_tuple("Cool-Warm (diverging)", ColormapAdapter<257>(CoolWarmColorMap))};
+    std::make_tuple("Cool-Warm (diverging)", ColormapAdapter<257>(CoolWarmColorMap)),
+    std::make_tuple("AccentMap (qualitative)", ColormapAdapter<8, true>(AccentMap)),
```
schulzch: This one is just called "Accent" and the amount of classes should be included, e.g. "5-class Accent".

plugins/gui/src/TransferFunctionEditor.cpp
```diff
@@ -92,7 +93,12 @@ template <size_t PaletteSize> PresetGenerator ColormapAdapter(const float palett
             // Linear interpolation from palette.
             size_t i0 = static_cast<size_t>(std::floor(t * LastIndex));
             size_t i1 = static_cast<size_t>(std::ceil(t * LastIndex));
-            double it = std::fmod(t * LastIndex, LastIndex);
```
schulzch, line 95: Nice catch! 👍

plugins/gui/src/TransferFunctionEditor.cpp
```diff
@@ -133,8 +139,9 @@ void RainbowAdapter(param::TransferFunctionParam::TFNodeType& nodes, size_t n) {
     }
 }
 
-std::array<std::tuple<std::string, PresetGenerator>, 12> PRESETS = {
-    std::make_tuple("Select...", [](auto& nodes, auto n) {}), std::make_tuple("Ramp", RampAdapter),
+std::array<std::tuple<std::string, PresetGenerator>, 20> PRESETS = {
```
schulzch, line 142: Now that we have more and more color maps, this tuple could/should be split into a struct, e.g..:
```
struct Preset {
	std::string name;
	enum class Scale {
		Sequential,
		Diverging,
		Qualitative
	} scale;
	std::string notice; // harmful, de-pinked, ...
	std::string reference; // ColorBrewer, Paper, ...
};
```
(just an idea)

schulzch (changes requested):
Adding the amount of supported classes is important. Rest is optional

reinago (approved):

reinago (approved):

invor (approved):

schulzch (approved):

PR #571, opened 2020-06-22
Using spdlog as new log backend

gralkapk:


gralkapk:
Cuda compile issue fixed

vislib/src/sys/Log.cpp
```diff
@@ -97,96 +96,41 @@ void vislib::sys::Log::DebugOutputTarget::Msg(UINT level,
...
-#endif /* _WIN32 */
-
-    if (this->stream != NULL) {
-        this->filename = path;
-    } else {
-        this->filename.Clear();
-    }
-
+        : Target(level)/*, stream(NULL)*/ {
+    std::wstring_convert<std::codecvt_utf8<wchar_t>> cvt;
```
straubar: std::wstring_convert is deprecated in C++17.
codecvt_utf8<wchar_t> seems to be Windows-only (c.f. table on https://en.cppreference.com/w/cpp/locale/wstring_convert).

vislib/include/vislib/sys/Log.h
```diff
@@ -196,6 +208,7 @@ namespace sys {
              * @param path The path to the physical log file
              * @param level The log level used for this target
              */
+            [[deprecated("use char* version instead")]]
```
straubar: Why not just delete the function if it throws an error anyways? This way we can get rid of all calls to this function instead of finding out when MegaMol explodes.

vislib/include/vislib/sys/Log.h
```diff
@@ -20,7 +20,14 @@
 #include "vislib/StringConverter.h"
 #include <cstdio>
 #include <ctime>
+#include <string>
+#include <iostream>
+#include <codecvt>
```
straubar, line 25: `codecvt` is not used anymore and can be removed.

straubar (commented):
Looks good so far, have not tested it though.

PR #572, opened 2020-06-22
start of Accelerating Parallel Coordinates thesis

invor:


moritz-h:
Squashed for merge in #777

PR #592, opened 2020-07-28
Updating glowl version

invor:
Updating glowl to v0.4a. Required some updates in MegaMol code due to breaking changes but seems to work for now. Could probably be merged once glowl version moves from alpha to beta (i.e. 0.4b), which will feature a more thorough use of direct state access.


braunms (approved):
Approving gui and cinematic plugin changes.
A only quick infovis flag storage test seemed to work, too.

PR #603, opened 2020-07-29
Added compile define for static spdlog

straubar:
The current external is used as a header-only library. Adding the compile definition, it is now used as a static library.
Also, spdlog was updated to the latest stable version (1.7.0).

(This is was initially thought as a try to reduce the compile time of MegaMol after spdlog was added.)


externals/CMakeExternals.cmake
```diff
@@ -484,6 +512,41 @@ function(require_external NAME)
...
+      GIT_REPOSITORY https://github.com/gabime/spdlog.git
+      GIT_TAG "v1.5.0"
+      DEPENDS fmt
+      BUILD_BYPRODUCTS "<INSTALL_DIR>/${SPDLOG_LIB}"
+      DEBUG_SUFFIX "d"
+      CMAKE_ARGS
+        -DSPDLOG_BUILD_EXAMPLE=OFF
+        -DSPDLOG_BUILD_TESTS=OFF
+        -DSPDLOG_FMT_EXTERNAL=ON
+        -DSPDLOG_COMPILED_LIB=ON
```
straubar: This is the added line of code (542).

gralkapk (approved):

PR #607, opened 2020-07-31
GUI - Add 3D Rotation Widget

braunms:
- plus: fixed positioning of GUI-Windows
- plus: fixed saving of Graph which has been loaded from core


braunms:
rot_widget

tobiasrau (approved):

PR #614, opened 2020-08-05
uchar is now also reinterpreted in adios containers

tobiasrau:


reinago (approved):
tests with the IFP hysolar laser scan looked OK to me

PR #620, opened 2020-09-03
Replacement of Shader Factory

gralkapk:


core/include/mmcore/utility/graphics/GLSLShader.h
```diff
@@ -0,0 +1,70 @@
...
+
+    GLint get_uniform_location(std::string const& name) { return _uniform_map[name].location; }
+
+    GLsizei get_uniform_size(std::string const& name) { return _uniform_map[name].count; }
+
+    void enable() const { glUseProgram(_program); }
+
+    void disable() const { glUseProgram(0); }
+
+    ~GLSLShader() { glDeleteShader(_program); }
```
geringsj: Is glDeleteShader correct here, or do we want to glDeleteProgram?

core/include/mmcore/utility/graphics/GLSLShader.h
```diff
@@ -0,0 +1,70 @@
...
+            auto name = std::make_unique<char[]>(max_name_len);
+
+            GLsizei length = 0;
+            GLsizei count = 0;
+            GLenum type = GL_NONE;
+
+            for (GLint idx = 0; idx < uniform_count; ++idx) {
+                glGetActiveUniform(_program, idx, max_name_len, &length, &count, &type, name.get());
+
+                _uniform_map.emplace(std::make_pair(std::string(name.get(), length),
```
geringsj: We could also carry along the uniform type here.

geringsj (approved):

geringsj (changes requested):
Fix https://github.com/UniStuttgart-VISUS/megamol/pull/620/files#r492100091

geringsj (dismissed):

moritz-h (changes requested):
Maybe we should discuss the following design choices before merging:
- Should we really name generic Shaders with `*.h` instead of `*.glsl`. Maybe `.glsl` is a better separation from C++ files, also syntax highlighting is may be better with glsl depending on editor.
- What is the purpose of the GLSLShader class within MegaMol? Do we want to remove this in favor of glowl `GLSLProgram`? And if not, why not also move this class to the library?

moritz-h (approved):

PR #630, opened 2020-10-06
use the png text fields for the project instead

reinago:
also, added some datetime formatting utility from vislib


reinago:
TODO: move the actual code to core::utils

braunms (approved):

PR #637, opened 2020-10-15
TableSelectionTx: Add param to use column as index

moritz-h:


plugins/mmstd_datatools/src/table/TableSelectionTx.cpp
```diff
@@ -23,6 +26,8 @@ TableSelectionTx::TableSelectionTx()
     , flagStorageReadOutSlot("readFlagStorageOut", "Flag storage read output")
     , flagStorageWriteOutSlot("writeFlagStorageOut", "Flag storage write output")
     , updateSelectionParam("updateSelectionParam", "Enable selection update")
+    , useColumnAsIndexParam("useColumnAsIndexParam", "Use column as index instead of row id")
+    , indexColumnParam("indexColumnParam", "Numeric index of column, which is used as row index")
```
reinago: why do the display names end in "param"?

reinago (approved):

PR #643, opened 2020-10-21
New modules for table data

gralkapk:
- Writer for CSV data
- Module that splits a table in two parts based on substring search in column names


schulzch:
As far as I know, `std::to_string` is depends on the `locale`. Rest looks good.

invor (approved):
Looking good. Survived my short test run (loading mmftd and writing back as csv).

PR #644, opened 2020-10-21
gui plugin

braunms:
see #539


invor:
Slider ist pretty nice for parameters with a given small range, but would it be possible to also have that interaction type with mouse click + drag right/left for a relative +/- offset (see e.g. backCol in View3D_2) ?

braunms:
win_size_pos

braunms:
> 
> 
> Slider ist pretty nice for parameters with a given small range, but would it be possible to also have that interaction type with mouse click + drag right/left for a relative +/- offset (see e.g. backCol in View3D_2) ?

You name it, you get it ...

invor (approved):

PR #668, opened 2020-11-23
Updated image call structure, loader and renderer

schatzkn:


reinago (commented):
TODO: playlists and stereo pairs (see renderer)

PR #670, opened 2020-11-26
gui plugin: bunch of fixes

braunms:
see #539


plugins/gui/src/GUIWindows.h
```diff
@@ -175,11 +175,11 @@ namespace gui {
 
         /** The global state (for settings to be applied before ImGui::Begin). */
         struct StateBuffer {
-            bool gui_pre_drawing_enabled;                  // Flag indicating whether GUI should be drawn
-            bool gui_post_drawing_enabled;                 // Prevent gui drwing change between pre and post draw
```
reinago, line 179: why do we need two separate bools here?

braunms: Yes, this is required to prevent changes to 'gui_enabled' being applied between calling pre and post drawing function. ImGui could be crashed this way ... Clarified in commit 6ff9cb3

reinago (approved):

Issue #671, opened 2020-11-26
Compositing_gl improvements


invor:
Feature requests, bugs and UX discussion for compositing_gl plugin.

# Features

- [ ] Add depth of field effect, e.g. following "Practical Gather-based Bokeh Depth of Field" chapter in GPU Zen
- [x] Performance/quality improvements for SSAO, e.g. based on https://gpuopen.com/fidelityfx-cacao/
- [x] Better anti-aliasing for real-time rendering, e.g. based on http://www.iryoku.com/smaa/
- [ ] Additional lighting BRDFs (and a parameter to change it)
   - [ ] Phong brdf (in progress)
   - [ ] PBR brdfs?
   - [x] Non-photorealistic shading, e.g. cartoon shading (in planning, steal cartoon shaders from protein?)
- [ ] MatCap shading mode
- [ ] Support for additional light sources in LocalLighting module
   - [ ] Area lights
   - [ ] HDRI
- [ ] Add option for selecting texture layout of module outputs (e.g. for TextureCombine). Right now its hardcoded to output RGBA16F
- [ ] Normal smoothing, e.g. https://github.com/UniStuttgart-VISUS/megamol/blob/23169c3d17e4ccc7b75bb8e7c29179c06c2dd881/plugins/mmstd_moldyn/Shaders/mipdepth6.btf#L610 i.e. https://onlinelibrary.wiley.com/doi/abs/10.1111/j.1467-8659.2009.01698.x

# Bugs/Improvements

- [ ] Split modules if needed/useful: Effects currently share a module that has all parameters and no way to hide parameters/slot based on actually active mode

# Feedback

- [ ] ...

PR #675, opened 2020-12-07
Compositing Updates 

invor:
Will add some of the features dicussed in #671


straubar (approved):
Tested local lighting with new depth slot.

PR #684, opened 2021-01-28
Refactor MMFTDataSource + add Reload Button

moritz-h:
- add data reload button
- removed VISLib, use std::ifstream
- cleanup

### DIFF
--- a/plugins/mmstd_datatools/src/table/MMFTDataSource.cpp
+++ b/plugins/mmstd_datatools/src/table/MMFTDataSource.cpp
@@ -10,12 +10,41 @@
 #include "mmcore/param/ButtonParam.h"
 #include "mmcore/param/FilePathParam.h"
 
-#include "vislib/String.h"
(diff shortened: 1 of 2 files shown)


plugins/mmstd_datatools/src/table/MMFTDataSource.cpp
```diff
@@ -1,145 +1,179 @@
...
-            : TableDataCall::ColumnType::QUANTITATIVE);
-        float f;
-        ASSERT_READ(&f, 4);
-        ci.SetMinimumValue(f);
-        ASSERT_READ(&f, 4);
-        ci.SetMaximumValue(f);
-    }
+    try {
+        std::string magicId = read_string(file, 6);
+        if (magicId != "MMFTD") {
```
gralkapk: Will be removed in C++20. Use `compare` instead.
Should also be able to cope with following `\0` -> no need to delete them.

gralkapk (approved):

PR #691, opened 2021-02-01
Pull Request Templates

### LINKED ISSUE #690: RFC: Pull Request Templates
Opened a PR including PR templates to help with the review process.
One template requests specific test instructions.
One template does not have that. Without it, a formal test cannot be guaranteed to be a part of the review.
However, this should be sufficient for smaller PRs or PRs that add plugin functionality and do not interfere with core functionality.

Open for discussion on the template format.

No clue if the issue header helps Github to interpret the PR template.

Backreference to PR #691

gralkapk:
Referencing Issue #690

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/require_test.md
+++ b/.github/PULL_REQUEST_TEMPLATE/require_test.md
@@ -1,0 +1,24 @@
+---
+name: PR with Test Requirements
+about: Create a PR with explicit test instructions.
+title: ''
+labels: needs testing
+assignees: ''
+
+---
+<!--Add a description here.
+Include a list of issues it fixes in the "Fixes #[]" format if applicable.-->
+
+## Summary of Changes
+<!--A list of changes that will be copied into the merge commit message and changelog.-->
+
+## References
+<!--A list of references of this PR, for instance related PRs or scientific papers.-->
+
+## Additional context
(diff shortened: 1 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE/require_test.md
```diff
@@ -0,0 +1,24 @@
+---
+name: PR with Test Requirements
+about: Create a PR with explicit test instructions.
+title: ''
+labels: needs testing
```
moritz-h, line 5: Is this a space separated list? May this should then be within ''

moritz-h:
Templates are nice.
I'm just thinking if the points 'References' and 'Additional context' are needed. What are you expecting the user should write here? I would expect that this two points are empty in > 90% of PRs, and therefore may should not be in the template.

If a reference is needed, may markdown links are used directly without a separate section.
The additional context section sounds at first look like a placeholder, here you can write whatever does not fit within the template. I think as it is just a template and the user can edit everything on special requirements, and therefore we may not need a special placeholder for this kind of stuff.

gralkapk:
If you have nothing to write in "References" and "Additional Context" then leave them empty.
The purpose of a template is to provide a common structure.

reinago (approved):

PR #701, opened 2021-02-08
CUDA service

gralkapk:


frontend/services/cuda/CUDA_Service.cpp
```diff
@@ -0,0 +1,46 @@
...
+    megamol::core::utility::log::Log::DefaultLog.WriteInfo(msg.c_str());
+}
+
+
+bool megamol::frontend::CUDA_Service::init(void* configPtr) {
+    cuInit(0);
+    int deviceCount = 0;
+    cuDeviceGetCount(&deviceCount);
+    if (deviceCount == 0) {
+        throw std::runtime_error("[CUDA_Service]: No Cuda device available");
```
geringsj: Instead of exception: log error and return 0
This way the frontend can shut down other services gracefully before terminating MegaMol. 
There is nobody waiting to catch this exception, anyway.

geringsj (dismissed):

frontend/services/cuda/CUDA_Service.cpp
```diff
@@ -0,0 +1,59 @@
...
+        return false;
+    }
+    int deviceCount = 0;
+    cuDeviceGetCount(&deviceCount);
+    if (deviceCount == 0) {
+        log_error("No Cuda device available");
+        return false;
+    }
+    cuDeviceGet(&ctx_.device_, 0);
+    auto ctx_ptr = reinterpret_cast<CUcontext*>(&ctx_);
```
geringsj: I guess **&ctx_** should be **&ctx_.ctx_**

geringsj (approved):

PR #722, opened 2021-02-18
GUI State Resource

braunms:
GUI Plugin: Bug and Feature Tracker #539

=> Depends on changes in PR #733.


frontend/main/src/main.cpp
```diff
@@ -244,7 +244,7 @@ int main(int argc, char* argv[]) {
         std::string result;
         if (megamol::core::utility::graphics::ScreenShotComments::EndsWithCaseInsensitive(file, ".png")) {
             if (!lua_api.RunString(
-                    megamol::core::utility::graphics::ScreenShotComments::GetProjectFromPNG(file), result)) {
+                    megamol::core::utility::graphics::ScreenShotComments::GetProjectFromPNG(file), result, file)) {
```
braunms: Also pass file name of screenshot. This is not further used within lua, but sets the lua ScriptPath. The script path is used in GUIWindows to recognise when another project has been loaded via lua. This changed script path is used to load the GUI state from the new project file.

frontend/services/gui/GUI_Service.cpp
```diff
@@ -188,7 +191,7 @@ void GUI_Service::digestChangedRequestedResources() {
 
 void GUI_Service::resetProvidedResources() {
 
-    // nothing to do here.
+    this->m_providedResource.request_gui_state = nullptr;
```
geringsj: Restting resources is intended to clean up resource state after a frame, before new resource state is written in the next frame. For exammple, keyboars events are deleted from event queues during resource reset and before updating the queues with new inputs. After a reset resources should be in a valid state. `request_gui_state` is not a resource that needs resetting.

plugins/gui/src/GUIWindows.cpp
```diff
@@ -996,24 +996,30 @@ bool megamol::gui::GUIWindows::SynchronizeGraphs(megamol::core::MegaMolGraph* me
...
             if (!this->state.project_script_paths.empty()) {
                 script_filename = this->state.project_script_paths.front();
             }
             // Load GUI state from project file when project file changed
             if (script_filename != this->state.last_script_filename) {
                 graph_ptr->SetFilename(script_filename);
-                this->load_state_from_file(graph_ptr->GetFilename());
                 this->state.last_script_filename = script_filename;
+
+                if (megamol::core::utility::FileUtils::FileHasExtension(script_filename, ".png")) {
```
geringsj: We now have 3 different places in three different services where this code is used (copy-pasted). Need to centralize project loading ASAP. Another service?

frontend/services/gui/GUI_Service.cpp
braunms: Jepp ok, in fact I had some headache about that ;-) Fixed in commit 00fe89a

frontend/resources/include/GUI_Resource.h
```diff
@@ -0,0 +1,20 @@
...
+
+#pragma once
+
+namespace megamol {
+namespace frontend_resources {
+
+struct GUIResource {
+
+    // Function requesting updated GUI state (e.g. ScreenshotService save project with GUI state tp PNG header)
+    std::function<std::string(void)> request_gui_state = nullptr;
```
geringsj: I think it is easier to have a function that always returns something, i.e. an empty string could also be a valid GUI state. Then nobody needs to check for a null pointer.

geringsj:
Centralized loading of Lua Projects and Screenshots can now be done using the ProjectLoader resource that comes with PR #733

reinago (approved):

PR #744, opened 2021-03-10
A camera reborn

invor:
Replaces Camera_2 with a smaller, more concise camera implementation. Reworks some concepts of views and camera handling in general. Further updates to rendering call infrastructure.

## Summary of Changes
- [x] New camera class added
- [x] Old camera class removed
- [x] Actual support for both perspective and orthographic projections
- [x] Support for cameras with given matrices and "unknown" intrinsics
- [x] Reworked interface between views and frontends
- [x] Support for overriding camera (controls) in view by providing camera in frontend
- [x] Replaces vislib framebuffer object with glowl framebuffer object implementation
- [ ] All existing plugins ported and working
- [ ] Support for tiled displays in Views and Camera
- ...

## Known Bugs
- SplitView seperator not visible
- Separate OpenGL stuff (e.g. view cube) drawn in GUI is not considered in any final fbo

## Known hacks that need to be cleaned up
- [x] GUI is re-drawn in resetProvidedResources, cause its overdrawn by PresentRenderedImages otherwise (commented redundant ImGui draw call in GUIManager::PostDraw)
- [ ] default framebuffer is "manually" set in main.cpp before postGraphRender to ensure everything that happens post-draw ends up in default window fbo (hence included glad.h in main.cpp)
- [ ] Camera aspect ratio is fixed ad-hoc in OnRenderView for SplitView use-cases

## Test Instructions
Test everything.


core/src/view/View3DGL.cpp
```diff
@@ -239,8 +226,26 @@ bool view::View3DGL::OnMouseButton(view::MouseButton button, view::MouseButtonAc
...
+        wndHeight = static_cast<int>(static_cast<float>(_fbo->GetHeight()) / tile_size.y);
+    } else if (projType == Camera::ProjectionType::PERSPECTIVE) {
+        auto tile_end = _camera.get<Camera::OrthographicParameters>().image_plane_tile.tile_end;
+        auto tile_start = _camera.get<Camera::OrthographicParameters>().image_plane_tile.tile_start;
+        auto tile_size = tile_end - tile_start;
+
+        wndWidth = static_cast<int>(static_cast<float>(_fbo->GetWidth()) / tile_size.x);
+        wndHeight = static_cast<int>(static_cast<float>(_fbo->GetHeight()) / tile_size.y);
+    } else {
+        return false; // Oh bother...
```
geringsj:
```
auto tile_end = _camera.get<Camera::PerspectiveParameters>().image_plane_tile.tile_end;
auto tile_start = _camera.get<Camera::PerspectiveParameters>().image_plane_tile.tile_start;
auto tile_size = tile_end - tile_start;
wndWidth = static_cast<int>(static_cast<float>(_fbo->GetWidth()) / tile_size.x);
wndHeight = static_cast<int>(static_cast<float>(_fbo->GetHeight()) / tile_size.y);
```

Is there some place where this can go as a helper function, e.g. into the image_tile? Because the same code is also used in View3DGL::OnMouseMove and might be useful in other Views too?

core/src/view/View2DGL.cpp
```diff
@@ -179,12 +179,17 @@ void view::View2DGL::ResetView(void) {
...
+
+        glBindFramebuffer(GL_FRAMEBUFFER, 0); // better safe then sorry, "unbind" fbo before delting one
+        try {
+            _fbo = std::make_shared<glowl::FramebufferObject>(width, height);
+            _fbo->createColorAttachment(GL_RGBA8, GL_RGBA, GL_UNSIGNED_BYTE);
+
+            // TODO: check completness and throw if not?
+        } catch (glowl::FramebufferObjectException const& exc) {
+            megamol::core::utility::log::Log::DefaultLog.WriteError(
+                "[View3DGL] Unable to create framebuffer object: %s\n", exc.what());
```
geringsj: The warning should warn about failing in View2DGL (says 3DGL)

geringsj:
Suggesting to cherry-pick https://github.com/UniStuttgart-VISUS/megamol/commit/9f53685b7bae6b5d220df0b9b02afee6a476da02 and https://github.com/UniStuttgart-VISUS/megamol/commit/67e12e42c19d27b5f61fa21b38235fdc5f4a5ecd from [tiling_experiments](https://github.com/UniStuttgart-VISUS/megamol/commits/tiling_experiments) as further Camera improvements

core/include/mmcore/utility/SDFFont.h
```diff
@@ -464,8 +462,9 @@ namespace utility {
         * @param a The rotation angle in degrees.
         * @param v The rotation axis.
         */
+        /// TODO XXX FIXME Currently not working ...
         inline void SetRotation(float a, glm::vec3 v) {
-            this->rotation = glm::quat((a * (float)M_PI / 180.0f), v.x, v.y, v.z);
+            this->rotation = glm::quat(glm::radians(a), v.x, v.y, v.z);
```
geringsj, line 467: how important is this? should we fix this?

core/include/mmcore/view/AbstractView.h
```diff
@@ -115,56 +116,43 @@ class MEGAMOLCORE_API AbstractView : public Module, public megamol::frontend_res
...
      *
-     * @param context The context information like time or GPU affinity.
+     * @param time ...
+     * @param instanceTime ...
      */
-    virtual void Render(const mmcRenderViewContext& context, Call* call) = 0;
+    using ImageWrapper = megamol::frontend_resources::ImageWrapper;
+    virtual ImageWrapper Render(double time, double instanceTime) = 0;
+
+    virtual ImageWrapper GetRenderingResult() const = 0;
```
geringsj, line 141: If `Render()` already returns the Image we dont need `GetRenderingResult()` anymore?

core/include/mmcore/view/BaseView.h
```diff
@@ -0,0 +1,410 @@
...
+            }
+
+            if (this->_cameraIsMutable) {
+                // get window resolution to help computing mouse coordinates
+                int wndWidth;
+                int wndHeight;
+                auto projType = this->_camera.template get<Camera::ProjectionType>();
+                if (projType == Camera::ProjectionType::PERSPECTIVE) {
+                    auto tile_end = this->_camera.template get<Camera::PerspectiveParameters>().image_plane_tile.tile_end;
+                    auto tile_start = this->_camera.template get<Camera::PerspectiveParameters>().image_plane_tile.tile_start;
```
geringsj, line 316: The `wndWidth` and `wndHeight` computation looks somewhat copy-pasted here and in `OnMouseMove()`. Can we minimize room for error by re-using lambdas or something like that?

core/include/mmcore/view/View3D.h
```diff
@@ -7,13 +7,23 @@
...
 namespace megamol {
 namespace core {
 namespace view {
 
-class MEGAMOLCORE_API View3D : public view::AbstractView3D {
+    inline constexpr auto cpu_fbo_resize = [](std::shared_ptr<CPUFramebuffer>& fbo, int width,
+                                              int height) -> void {
+        fbo->width = width;
+        fbo->height = height;
+        // TODO reallocate buffer?
```
geringsj, line 23: This looks important?

frontend/main/src/main.cpp
```diff
@@ -223,9 +225,13 @@ int main(const int argc, const char** argv) {
 
             imagepresentation_service.RenderNextFrame(); // executes graph views, those digest input events like keyboard/mouse, then render
 
+            /// XXX temporary fix to make sure that everything that happens post-draw ends up in default window fbo...
+            glBindFramebuffer(GL_FRAMEBUFFER, 0);
```
geringsj, line 229: Where is the right place to actually fix this? Image Presentation?

frontend/services/gui/src/GUI_Service.cpp
```diff
@@ -389,6 +392,14 @@ namespace frontend {
     }
 
 
+    void GUI_Service::resource_provide_gui_render() {
+
+        if (this->m_gui != nullptr) {
+            this->m_gui->DrawUiToScreen();
+        }
+    }
```
geringsj, line 400: I believe in the long run we want such a GUI-Render callback to be called/registered by Image Presentation. Question is if this current callback hack can stay or should be reworked?

frontend/services/image_presentation/ImagePresentation_Service.cpp
```diff
@@ -219,5 +244,30 @@ bool ImagePresentation_Service::clear_entry_points() {
     return true;
 }
 
+void ImagePresentation_Service::present_images_to_glfw_window(std::vector<ImageWrapper> const& images) {
```
geringsj, line 247: TODO: set GL default framebuffer here?

frontend/services/image_presentation/ImagePresentation_Sinks.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+void glfw_window_blit::set_framebuffer_size(unsigned int width, unsigned int height) {
+    fbo_width = width;
+    fbo_height = height;
+}
+
+void glfw_window_blit::blit_texture(unsigned int gl_texture_handle, unsigned int texture_width, unsigned int texture_height) {
+    // credit goes to: https://stackoverflow.com/questions/31482816/opengl-is-there-an-easier-way-to-fill-window-with-a-texture-instead-using-vbo
+
+    glNamedFramebufferTexture(gl_fbo_handle, GL_COLOR_ATTACHMENT0, gl_texture_handle, 0);
```
geringsj, line 38: TODO: named GL funcs only available with higher GL versions. need alternative if GL version does not provide such functions?

plugins/protein/src/UnstructuredGridRenderer.cpp
```diff
@@ -236,11 +234,8 @@ bool protein::UnstructuredGridRenderer::Render(core::view::CallRender3DGL& call)
 
     glScalef( scale, scale, scale);
 
-    float viewportStuff[4] = {
-        cameraInfo.image_tile().left(),
-        cameraInfo.image_tile().bottom(),
-        cameraInfo.image_tile().width(),
-        cameraInfo.image_tile().height()};
+    // TODO double-check whether this renderer is ever used with viewport != full fbo
+    float viewportStuff[4] = { 0, 0, cr3d->GetFramebuffer()->getWidth(), cr3d->GetFramebuffer()->getHeight()};
```
geringsj, line 238: Some day we may want to do that.

plugins/remote/src/RendernodeView.cpp
```diff
@@ -294,12 +294,9 @@ bool megamol::remote::RendernodeView::OnRenderView(core::Call& call) {
...
-    mmcRenderViewContext context;
-    ::ZeroMemory(&context, sizeof(context));
-    context.Time = time;
-    context.InstanceTime = crv->InstanceTime();
+    auto instanceTime = crv->InstanceTime();
 
-    this->Render(context, &call);
+    this->Render(time, instanceTime);
 
     return true;
```
geringsj, line 301: `Render()` now returns an ImageWrapper

plugins/remote/src/RendernodeView.h
```diff
@@ -69,7 +69,7 @@ class RendernodeView : public core::view::AbstractTileView {
      */
     virtual ~RendernodeView(void);
 
-    void Render(const mmcRenderViewContext& context, core::Call* call) override;
+    void Render(double time, double instanceTime) override;
```
geringsj, line 72: `Render()` now returns an ImageWrapper

geringsj (approved):
Should be fine i guess

core/include/mmcore/view/AbstractView.h
invor: I think I kept this around in case someone needs to access the rendering result after the call to `Render()` was already made elsewhere in the code, e.g. when a view is called by another view via Call callback. But at the moment I don't thinks it's possible for one view to directly call a function with return value of another view anyway...

core/include/mmcore/view/View3D.h
invor: It sure does, but it should be based on how CPU framebuffers were handled in master at this point and I think Ospray for example allocates its own memory and then overwrites the contents of the CPUFramebuffer.
So right now, this only tries to at least let the framebuffer know its size so that this information can already be used while the actual memory is provided by the renderer...

frontend/main/src/main.cpp
invor: I guess so, but might be needed in several places in the frontend that at the moment assume the OpenGL framebuffer state to be in default after returning from graph exec

frontend/services/image_presentation/ImagePresentation_Service.cpp
invor: +1

frontend/services/image_presentation/ImagePresentation_Sinks.cpp
invor: Is not having a high GL version a realistic scenario? Going bindless feel so liberating :D No state consequences (that I know of).

PR #755, opened 2021-03-17
Collect View Renderings in Frontend, Screenshots independend of Window size

geringsj:
* Reorganize management of Graph Entry Points in Frontend around ImagePresentation Service - collects rendering results from views in textures
* Proof of Concept: View3DGL and View2DGL output rendering result as `ImageWrapper` object to ImagePresentation Service
* CLI option to show images held by frontend in GUI via `--entryfbos`
* introduces `--framebuffer` and `--tile` options
* `mmSetFramebufferSize()` resizes all views independent of size of window
* `mmScreenshot()` writes frontend images as screenshots
* Small code refactorings here and there
* Fixes critical recursion bug when mmRenderNextFrame() occurs in project loaded via file-drop into the GLFW window
* Depends on PR #748
* Prepares frontend for PR #744 

## Summary of Changes
Pull MegaMolGraph Entry Points into separate Service - the Image Presentation -, give each Graph Entry Point a Texture that the corresponding View writes rendering results into, and thus enable rendering and screenshots independent of GLFW Window size. In the long term, enables sophisticated Split-View like layouting of View results in the ImGui or using some other layouting mechanism. 

## References and Context
Depends on refactoring of all relevant Views to export their FBO. The Frontend<->View interface minimizes data copying by simply wrapping the texture object of the view using a `ImageWrapper` object, which takes a non-owning reference to the wrapped texture. Currently, only View3DGL and View2DGL are hooked into the frontend image mechanism, but connecting more views should not be too difficult (see `GetRenderingResult()` in `AbstractView.h`, `View3DGL.h` and `View2DGL.h`).

The frontend images are designed as zero-copy wrappers (`ImageWrapper`) and are able to generalize over Views that render images into CPU Buffers and into OpenGL Textures. Actually using contents of an `ImageWrapper` means casting it into either a `gl_texture` or `byte_texture` object, which then allows API-specific access to texture contents. This generic wrapping and separation of API details allows to focus API usage (e.g. OpenGL) in a few select places, which then can be turned off more easily when the API is not supported due to machine restrictions (OpenGL in HPC rendering).

The code is ready to write screenshots of an entry point view when given the view name, but the Lua and GUI screenshot APIs are not yet adjusted to handle this. Until we can manage 'optional' arguments in Lua Lambda Callbacks, `mmScreenshot()` writes a screenshot for each View that is currently a graph entry point.

## Test Instructions
Load a MegaMol Lua Project that contains View3DGL or View2DGL as entry points with CLI option --entryfbos. Take screenshots via GUI or `mmScreenshot()`. Set framebuffer size via `--framebuffer` or `mmSetFramebufferSize()`. Tiling support is somewhat present, but needs to be fleshed out depending on tiling support of the new camera.


frontend/resources/src/GenericImage.cpp
```diff
@@ -0,0 +1,515 @@
...
+#include <cstring>
+#include <list>
+
+using namespace megamol::frontend_resources;
+
+static void init_texture(unsigned int& handle) {
+    int old_handle = 0;
+    glGetIntegerv(GL_TEXTURE_BINDING_2D, &old_handle);
+
+    glCreateTextures(GL_TEXTURE_2D, 1, &handle);
```
invor: Does glCreateTextures modify the current texture binding, i.e., is there really a need to get and re-set the binding?

frontend/resources/include/GenericImage.h
```diff
@@ -0,0 +1,160 @@
...
+    using byte = unsigned char;
+    // if byte image, image_data holds (width * height * channels) entries.
+    // row-major image starting at bottom-left pixel
+    // as in https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glReadPixels.xhtml
+    mutable std::vector<byte> image_data; // mutable for texture download
+
+    // note that gl.h defines "typedef unsigned int GLuint;"
+    mutable unsigned int gl_texture_handle = 0; // mutable for texture upload
+    void download_texture() const; // downloads texture into local image_data
+    void upload_texture() const; // downloads texture into local image_data
```
invor: Guess the comment is not accurate? This one uploads texture data from local image data, right?

invor (dismissed):
Looks (and runs) fine on my end, but I'm not very familiar with the frontend (services) so I concentrated on looking through the GenericImage, ImagePresentation and changes to the Views.

As discussed before, we will have to figure out the details and final interface between Views and ImagePresentationService once we get all of this into #744, because this is as good as it gets without the new camera and further major updates to the Views.

frontend/resources/src/GenericImage.cpp
geringsj: Removed restoring old handle. It didnt do anything.

geringsj:
> As discussed before, we will have to figure out the details and final interface between Views and ImagePresentationService once we get all of this into #744, because this is as good as it gets without the new camera and further major updates to the Views.

Rewrote the frontend images to hold a void* reference to passed texture objects. When somebody actually wants to use the wrapped texture object returned by views, he needs to cast it to a `gl_texture` or `byte_texture` object. This way we move all the API-specific texture handling to the place where it is actually used and understood.

PR #757, opened 2021-03-23
SDF Font

braunms:
Refactoring.

## Summary of Changes
- Replaced use of legacy OpenGL matrix stack with explicit matrices for DrawString functions.
- Extended outlining mode.
- Some renaming, clean up, updated comments ...

Fixed usage for modules in infovis and cinematic plugin.

## Test Instructions
E.g. font rendering is used in infovis.lua or cinematic_editor.lua examples.


invor (dismissed):
Looking good overall. Infovis example (which uses fonts) worked without issues on AMD GPU. One suggestion: It might be possible to save quite a few lines of code for the constructors of the SDFFont class (SDFFont.cpp) by just writing out the one with all arguments and have the other ones call that with their varying default values.

braunms:
Thanks, good point. Took care of useless ctor code in e359056.

reinago (approved):

PR #763, opened 2021-03-29
gui mouse cursor changes via glfw

braunms:
Enabled mouse cursor shape changes via glfw. This access to the 'backend' for imgui is required to enable functions that require non-standard mouse cursor shapes. This enables the option to change the window size via edges.

## Summary of Changes
Adapted code from imgui_impl_glfw.cpp. to generate different mouse cursors for GUI in MegaMol.


braunms:
@geringsj Maybe this can/should be moved to the frontend?

geringsj:
> @geringsj Maybe this can/should be moved to the frontend?

Wrapping the appropriate GLFW calls in WindowManipulation.h seems reasonable. This way we can avoid including and linking against GLFW in the GUI, as the OpenGL GLFW Service should isolate the rest of MegaMol from GLFW. Or else when someone wants/needs to replace GLFW for something else he will run into trouble (e.g. ImGui software rendering, running MegaMol with externally defined OpenGL Context source).

frontend/services/opengl_glfw/OpenGL_GLFW_Service.cpp
```diff
@@ -792,6 +804,47 @@ void OpenGL_GLFW_Service::glfw_onPathDrop_func(const int path_count, const char*
...
+    m_pimpl->mouse_cursors[2] = ::glfwCreateStandardCursor(GLFW_ARROW_CURSOR); // ImGuiMouseCursor_ResizeAll
+    m_pimpl->mouse_cursors[5] = ::glfwCreateStandardCursor(GLFW_ARROW_CURSOR); // ImGuiMouseCursor_ResizeNESW
+    m_pimpl->mouse_cursors[6] = ::glfwCreateStandardCursor(GLFW_ARROW_CURSOR); // ImGuiMouseCursor_ResizeNWSE
+    m_pimpl->mouse_cursors[8] = ::glfwCreateStandardCursor(GLFW_ARROW_CURSOR); // ImGuiMouseCursor_NotAllowed
+#endif
+    ::glfwSetErrorCallback(prev_error_callback);
+}
+
+void OpenGL_GLFW_Service::update_glfw_mouse_cursor(const int cursor_id) {
+    if (!m_pimpl->config.windowPlacement.noCursor && (cursor_id < m_pimpl->mouse_cursors.size())) {
```
geringsj: Should we check `cursor_id < 0` here? Because in `GetMouseCursor()` it says `return ImGui::GetMouseCursor(); // ImGuiMouseCursor_None = -1`. We could also change `cursor_id` to `unsigned int` and eliminate this possible error here by making somebody else cast/check their provided cursor id - e.g. `GetMouseCursor()`

geringsj (approved):

PR #774, opened 2021-04-07
Revamp file system storage of multiple different camera settings

schatzkn:
## Summary of Changes
- Revamped storing of multiple cameras via CTRL+Number


geringsj:
Camera and View3D have been revamped since.

PR #776, opened 2021-04-13
Frontend 1: ImagePresentation Service manages Graph Entry Points

geringsj:
Refactoring of PR #755. Depends on PR #775.
The start of this PR is commit 9c27a01 "add ImagePresentation Service: pokes rendering of graph entry points"

* Reorganize management of Graph Entry Points in Frontend around ImagePresentation Service - collects rendering results from views in textures
* Proof of Concept: View3DGL and View2DGL output rendering result as `ImageWrapper` object to ImagePresentation Service
* Depends on PR #748
* Prepares frontend for PR #744 

## Summary of Changes
Pull MegaMolGraph Entry Points into separate Service - the Image Presentation -, give each Graph Entry Point a Texture that the corresponding View writes rendering results into.
This enables rendering and screenshots independent of GLFW Window size and will be implemented in a later PR. In the long term, enables sophisticated Split-View like layouting of View results in the ImGui or using some other layouting mechanism. 

## References and Context
Depends on refactoring of all relevant Views to export their FBO. The Frontend<->View interface minimizes data copying by simply wrapping the texture object of the view using a `ImageWrapper` object, which takes a non-owning reference to the wrapped texture. Currently, only View3DGL and View2DGL are hooked into the frontend image mechanism, but connecting more views should not be too difficult (see `GetRenderingResult()` in `AbstractView.h`, `View3DGL.h` and `View2DGL.h`).

The frontend images are designed as zero-copy wrappers (`ImageWrapper`) and are able to generalize over Views that render images into CPU Buffers and into OpenGL Textures. 

## Test Instructions
Not immediately testable since this PR provides infrastructure for the Camera redesign. Make sure that all your important projects load and render correctly.

### DIFF
--- a/core/src/MegaMolGraph.cpp
+++ b/core/src/MegaMolGraph.cpp
@@ -4,7 +4,6 @@
 #include "mmcore/MegaMolGraph.h"
 
 #include "mmcore/AbstractSlot.h"
-#include "mmcore/view/AbstractView_EventConsumption.h"
 
 #include "mmcore/utility/log/Log.h"
 
@@ -156,6 +155,15 @@ bool megamol::core::MegaMolGraph::RenameModule(std::string const& oldId, std::st
         }
         if (matches_old_prefix(call.request.to)) {
             put_new_prefix(call.request.to);
+        }
+    }
+
+    // dont know what we are supposed to do when entry point renaming fails... how can it fail?
+    if (module_it->isGraphEntryPoint) {
+        bool view_rename_ok = m_image_presentation->rename_entry_point(oldId, newId);
+        if (!view_rename_ok) {
+            log_error("error renaming graph entry point. image presentation service could not rename module: " + oldId + " -> " + newId);
+            return false;
         }
     }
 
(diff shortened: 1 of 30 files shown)


core/src/MegaMolGraph.cpp
```diff
@@ -627,23 +625,51 @@ bool megamol::core::MegaMolGraph::RemoveGraphEntryPoint(std::string moduleName)
...
     }
 
-    this->graph_entry_points.remove_if([&](GraphEntryPoint& entry) { return entry.moduleName == moduleName; });
+    bool view_removal_ok = m_image_presentation->remove_entry_point(moduleName);
+
+    // note that for us it is not an error to try to remove a module as graph entry point
+    // if that module is not registered as an entry point
+    // but maybe image presentation may tell us that for some other reason removal failed?
+    if (!view_removal_ok) {
+        log_error("error adding graph entry point. image presentation service could not remove module: " + moduleName);
```
reinago, line 634: adding -> removing

core/src/MegaMolGraph.cpp
```diff
@@ -585,33 +586,30 @@ std::vector<megamol::core::param::AbstractParam*> megamol::core::MegaMolGraph::L
...
     }
     
-    auto module_ptr = module_it->modulePtr;
+    auto module_shared_ptr = module_it->modulePtr; // we cant cast shared_ptr to void* for image presentation rendering
+    auto& module_ref = *module_shared_ptr;
+    auto* module_raw_ptr = &module_ref;
 
-    auto resources = get_requested_resources(execution_resource_requests);
+    // the image presentation will issue the rendering and provide the view with resources for rendering
+    // probably we dont care or dont check wheter the same view is added as entry point multiple times
```
reinago, line 604: who knows, that might be a mechanism for stereo pairs, or it might not be

reinago (approved):

PR #777, opened 2021-04-13
Accelerating PCP BSc. Thesis

moritz-h:
Squash of BSc. Thesis Accelerating PCP.
Replaces: #572


invor (approved):
PCP still works fine. AmortizedRendering module still feel experimental, but that's okay.

PR #778, opened 2021-04-13
Frontend 2: ImagePresentation Presents View Renderings To GLFW Window

geringsj:
Depends on PR #744.
At the end of a frame, ImagePresentation blits (presents) the images output by Graph Entry Points (Views) into the GLFW window.

## Summary of Changes
* Move OpenGL Buffer Swap from OpenGL_GLFW Service to ImagePresentation Service
* ImagePresentation holds list of Presentation Sinks - those are callbacks that map `std::vector<ImageWrapper>` to some output (presentation sink)
* This allows for generalized registration/swapping of image outputs, e.g. when no OpenGL is available, write images to disk
* Experimental: draws GUI on top of view renderings

## References and Context
When PR #744 is finished, this PR enables generalized presentation of View rendering outputs in the frontend. The goal is that at some point Views dont render into the OpenGL default framebuffer anymore, but rather return a rendering result to the frontend, which can decide how to present that rendered image to the user.

## Test Instructions
For now, only works with projects containing views as entry points that pass their rendering result to the frontend, as introduced in #776. Currently only works for View3DGL and View2DGL. For example, start testspheres.lua. GUI ovrlay for presentation needs more thought put into it.


geringsj:
@braunms What would be the best approach to overlay the GUI over the presented view rendering results?

@invor Thoughts on general rendering performance and maintainability?

PR #780, opened 2021-04-16
transfer function - improvements

braunms:
Some minor improvements for transfer function (Parameter and Editor)

## Summary of Changes
- Fixed linear and gauss interpolation to calculate color value for center of texel
- Allow arbitrary range values (so far 'left' range value has to be smaller than 'right' value)
- Removed condition for transfer function to have always at least two nodes for values 0 and 1
- Better transfer function plot UX for nodes
- Added option to invert colors of transfer function
- Added node "paint mode" for plot
- Some variable renaming, changed log output, ...

**TODO:**
- [x] Improve existing node deletion in plot paint mode.
- [x] "Allow" nodes at 0 and 1 in plot paint mode. 

## References and Context
none

## Test Instructions
Start project containing transfer function module and open transfer function editor.


gralkapk (dismissed):
Would be nice if one could "paint" a TF.
This would come in handy especially if using preselected TFs.
They have no real nodes to adjust, for instance, the shape of the alpha curve.

reinago (approved):

PR #789, opened 2021-04-26
CLI Params: Set Graph Parameter Values via CLI

geringsj:
Introduces mechanism to pass Lua commands via CLI: `--execute "mmQuit()"`, `-e "mmQuit()"`
The CLI Lua commands are executed after project files.
Also introduces `--param param=value` CLI option to set Graph parameters via CLI.
`--param` options are executed before `--execute` options, but both are executed after project files loaded.

## Summary of Changes
* add `--execute` CLI option
* add `--param` CLI option
* in `main.cpp`, execute CLI lua commands after project loading

## Test Instructions
`./megamol.exe ../examples/testspheres.lua --param "::data::numspheres=42" --param "::view::anim::play=on"` shows 42 moving spheres
`./megamol.exe -e "mmQuit()"` immediately closes MegaMol


frontend/main/src/main.cpp
```diff
@@ -257,6 +257,17 @@ int main(const int argc, const char** argv) {
         }
     }
 
+    // execute Lua commands passed via CLI
+    if (graph_resources_ok)
+    if (!config.cli_execute_lua_commands.empty()) {
+        std::string lua_result;
+        // here we preserve the last script path, because the lua code we execute has no file name
+        run_megamol = lua_api.RunString(config.cli_execute_lua_commands, lua_result, lua_api.GetScriptPath());
```
reinago: you can pass an empty scriptPath since the LuaAPI will not accept "degrading" its knowledge anyway https://github.com/UniStuttgart-VISUS/megamol/blob/066b8ad9839ec0ee2e6941e07e7b974f78d68ce6/core/src/LuaAPI.cpp#L138-L140

reinago (approved):

PR #804, opened 2021-05-15
gui: parameter names

braunms:
Reworked parameter names in GUI graph.

## Summary of Changes
- parameter name consists of two separated parts: full name of parent module (group name + module name) and full name of parameter (parameter namespace/group + parameter name)
- fixed rendering of calls in configurator
- minor fix of parameter popup placement in configurator
- readded checkbox as widget alternative for bool params


straubar (approved):
- Modifying parameters of modules in a group now work correctly.
- Direct connections between collapsed groups are now shown.

PR #806, opened 2021-05-18
gui: additonal window registration

braunms:
Allow external callback registration for new GUI windows and pop-ups in frontend services.

## Test Instructions

See [developer information for GUI plugin](https://github.com/braunms/megamol/tree/gui/plugins/gui#gui-windowpopupnotification-for-frontend-service).


frontend/services/gui/GUI_Service.cpp
```diff
@@ -61,7 +60,8 @@ bool GUI_Service::init(const Config& config) {
         "ProjectLoader",                         // 8 - trigger loading of new running project
         "FrameStatistics",                       // 9 - current fps and ms value
         "RuntimeConfig",                         // 10 - resource paths
-        "WindowManipulation"                     // 11 - GLFW window pointer
+        "WindowManipulation",                    // 11 - GLFW window pointer
+        "ScreenshotGUIRenderRequest"             // 12 - gui render request from screenshot service
```
geringsj: It would be easier if the GUI Service does not need to know when other Services want to add a GUI window. For example the GUI Service could provide a "register GUI window callback" resource which other services can use to register ImGUI callbacks themselves.

frontend/services/gui/GUI_Service.cpp
```diff
@@ -247,6 +243,17 @@ void GUI_Service::digestChangedRequestedResources() {
     /// Get window manipulation resource = resource index 11
     auto& window_manulation = this->m_requestedResourceReferences[11].getResource<megamol::frontend_resources::WindowManipulation>();
     window_manulation.set_mouse_cursor(gui->GetMouseCursor());
+
+    // Register GUI windows only once
+    if (this->register_gui_callbacks_once) {
+
+        auto &screenshot_gui_render_request = this->m_requestedResourceReferences[12].getResource<megamol::frontend_resources::GUIRenderRequest>();
+        gui->RegisterGUIWindow(screenshot_gui_render_request.window_name, screenshot_gui_render_request.callback);
```
geringsj: Exposing a "Add GUI window callback" resource would also relieve the GUI Service from holding possible incoming render requests. Or the other way around: somebody who wants to register his GUI render request/callback does not need to change the GUI Service hpp/cpp to register his callback, if there is a resource that he can request and access in his own service.

frontend/services/screenshot_service/Screenshot_Service.hpp
```diff
@@ -63,6 +66,10 @@ class Screenshot_Service final : public AbstractFrontendService {
     std::vector<FrontendResource> m_providedResourceReferences;
     std::vector<std::string> m_requestedResourcesNames;
     std::vector<FrontendResource> m_requestedResourceReferences;
+
+    /// XXX TEMP
+    bool guiRenderCallback(WinConfig_t& wc);
+    megamol::frontend_resources::GUIRenderRequest m_guiRenderRequest_resource;
```
geringsj: It would be nice if we could simply register the GUI Render Request via a Lambda and not hold an additional resource in our service that we need to expose. For example, 
```
frontend_resources[gui_window_render_callbacks_resource].add_request( 
    {window_name, 
    [&](WinConfig_t& wc) {
    ... 
} } )
```

frontend/resources/include/GUIRenderRequest.h
```diff
@@ -0,0 +1,26 @@
...
+ * GUIRenderRequest.h
+ *
+ * Copyright (C) 2020 by VISUS (Universitaet Stuttgart).
+ * Alle Rechte vorbehalten.
+ */
+
+#pragma once
+
+#include "WindowCollection.h"
+typedef megamol::gui::WindowCollection::WindowConfiguration WinConfig_t;
```
geringsj: `WindowCollection.h` includes `GUIUtils.h` which includes 

```
#include "imgui.h"
#include "imgui_impl_opengl3.h"
#include "imgui_internal.h"
#include "imgui_stdlib.h"

#include "glm/glm.hpp"
#include "glm/gtc/matrix_transform.hpp"
#include "glm/gtc/type_ptr.hpp"
```

Is there a way to keep this resource minimal regarding the dependencies? If the Header files of Services and other things get blown up by such includes our compilation times may suffer. Is there a way to wrap `WinConfig_t` functionality locally and hide such header includes in a .cpp implementation file?
On the other hand, if somebody wants to use ImGui, it is kind of okay to pull in the necessary headers. Would be nice to somehow keep the headers clean though.

frontend/services/screenshot_service/Screenshot_Service.cpp
```diff
@@ -262,6 +275,20 @@ void Screenshot_Service::postGraphRender() {
     // swap buffers, glClear
 }
 
+/// XXX TEMP
+bool Screenshot_Service::guiRenderCallback(WinConfig_t& wc) {
+    bool valid_imgui_scope =
+            ((ImGui::GetCurrentContext() != nullptr) ? (ImGui::GetCurrentContext()->WithinFrameScope) : (false));
```
geringsj: Is there some way to automate this validity check, e.g. via a helper function one can call?

frontend/services/gui/GUI_Service.cpp
braunms: see commit b3fe3c4

frontend/services/screenshot_service/Screenshot_Service.hpp
braunms: see commit b3fe3c4

frontend/resources/include/GUIRenderRequest.h
braunms: see commit b3fe3c4

frontend/services/screenshot_service/Screenshot_Service.cpp
braunms: see commit b3fe3c4

frontend/services/gui/GUI_Service.cpp
braunms: see commit b3fe3c4

plugins/gui/src/WindowCollection.h
```diff
@@ -33,106 +41,99 @@ namespace gui {
...
+
+        void SetVolatileCallback(CallbackFunc_t* callback) {
+            this->volatile_callback = callback;
+            this->callback_id = DRAWCALLBACK_UNKNOWN;
+        }
+
+        std::string Name(void) const { return this->name; }
+        size_t Hash(void) const { return this->hash_id; }
+        PredefinedCallbackID CallbackID(void) const { return this->callback_id; }
+        CallbackFunc_t* Callback(void) const { return this->volatile_callback; }
```
geringsj: I am wondering: can we make the `CallbackFunc_t` owned here? Then Services requesting GUI windows could just push a lambda callback without holding a member variable for that callback themselves. This way we only need to include `GUIWindowRequest.h` in the Service cpp file. Also, then the GUI implementation could wrap its own `PredefinedCallbackID` callbacks in `CallbackFunc_t` via function pointers and this would simplify the code calling/rendering these callbacks?

plugins/gui/src/WindowCollection.h
```diff
@@ -33,106 +41,99 @@ namespace gui {
...
+        void ApplyWindowSizePosition(bool consider_menu);
+        
+        Complete config;
+
+    private:
+
+        size_t hash_id;                    // unique hash generated from name to omit string comparison
+        std::string name;                  // unique name of the window
+        PredefinedCallbackID callback_id;  // ID of the predefined callback drawing the window content
+        CallbackFunc_t* volatile_callback; // [NOT SAVED] Alternative unknown window callback.
```
geringsj: I am wondering: can we make the `CallbackFunc_t` owned here? Then Services requesting GUI windows could just push a lambda callback without holding a member variable for that callback themselves. This way we only need to include `GUIWindowRequest.h` in the Service cpp file. Also, then the GUI implementation could wrap its own `PredefinedCallbackID` callbacks in `CallbackFunc_t` via function pointers and this would simplify the code calling/rendering these callbacks?

geringsj (approved):

PR #817, opened 2021-06-07
CLI: load default config from executable location

geringsj:
Load default config from `megamol.exe` directory

### DIFF
--- a/frontend/main/src/CLIConfigParsing.cpp
+++ b/frontend/main/src/CLIConfigParsing.cpp
@@ -17,6 +17,30 @@ std::filesystem::path getHomeDir() {
 #endif
 }
 
+// find megamol executable path
+static
+std::filesystem::path getExecutableDirectory() {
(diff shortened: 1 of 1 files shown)


frontend/main/src/CLIConfigParsing.cpp
```diff
@@ -30,6 +30,29 @@ static std::string getHomeDir() {
 #endif
 }
 
+// find megamol executable path
+static std::string getExecutableDirectory() {
```
moritz-h: Why not just using std::filesystem:path as return type?

moritz-h (changes requested):
The #ifdef cascade to include filesystem is not needed anymore.
Just do `#include <filesystem>`. We require a modern enough compiler now anyway.

moritz-h (approved):

tobiasrau (approved):

PR #833, opened 2021-06-25
fixes to the bidirectional pattern in devguide

reinago:
main point: is this less convoluted/better to understand?


docs/devguide/devguide.md
```diff
@@ -98,19 +100,24 @@ To properly track changes across several Modules, you need to follow the recipe.
...
-- create a ```CallerSlot``` for ```DATACallWrite``` for modules supplying the ```DATA```
 
-A module with slots for both directions by convention should first execute the reading and then provide updates via writing.
+A module with slots for both directions by convention must, in this exact order:
+- execute the reading, overwriting its DATA with incoming update
+- apply logic that will modify the DATA (if parameters or user input cause this, for example)
+- provide DATA updates via writing.
 
 #### Usage: ```DATACallRead```
 - In the ```GetData``` callback, make sure you can *supply unchanged data very cheaply*.
```
gralkapk: This is meant for the holder, right? If so, mention it. The same for the case below.

docs/devguide/devguide.md
```diff
@@ -98,19 +100,24 @@ To properly track changes across several Modules, you need to follow the recipe.
 ### Recipe
 Split up the data flow for each direction, one call for reading only, one call for writing only.
 Keep in mind that the caller by definition is "left" in the module graph and the callee is "right". The callee is the end of a callback, but for this pattern this has nothing to do with the direction the data flows in, which results in the rules defined below.
+
+![bidir example](bidir.png)
+
 - set up a ```uint32_t version_DATA``` in each Module
```
gralkapk: Neither the consumer nor the producer should have their own version. This is a property of the DATA and should therefore always be "owned" by the holder. Imagine two producers A and B. A updates DATA before B. If B sets DATA with the version stored in B, it can happen that the set_version is lower than the get_version and hasUpdate() breaks.

gralkapk (changes requested):

docs/devguide/devguide.md
```diff
@@ -95,36 +98,52 @@ You can have a look for an example in the ```FlagStorage_GL``` together with the
...
+- create a ```CallerSlot``` for ```DATACallRead``` for *consumer*s
+- create a ```CallerSlot``` for ```DATACallWrite``` for *supplier*s
+
+#### Interaction with the ```DATA``` from outside:
+This must always follow the same order to avoid inconsistencies:
+- if *consumer*: execute the reading:
+  - issue the Call ```(*call)(DATACallRead::CallGetData)``` or your specialized version
+  - check if something new is available: ```call::hasUpdate()```
+  - overwrite/replace what notion you had of ```DATA``` if there is an incoming update
+  - take note of the ```call::version()``` *V* of the incoming update.
```
gralkapk, line 127: I don't think it is necessary to store V even temporarily. Since we cannot prevent race conditions with this version, we need to increase the version number of the call it has in the moment we update the data, if we want to communicate this update.

gralkapk (changes requested):

gralkapk (approved):

PR #834, opened 2021-06-30
Draw Histograms of Textures

reinago:
This is just an idea. I am not sure whether the implementation should be where it is. Also there is duplicated code ATM.
Histo bin number needs to be limited to 255 with the current implementation (better ideas?)


moritz-h:
I think I see a few problems with the current state of modularity/combined code/dependencies/etc., but of course wanting a histogram of anything other than a float table is a good idea. Maybe we can discuss in the next meeting about this, to not write that much text here :smiley:

moritz-h:
Related examples update is here: https://github.com/UniStuttgart-VISUS/megamol-examples/pull/9
Please merge both together.

moritz-h (dismissed):

moritz-h (approved):

PR #846, opened 2021-07-14
rework on file path param

braunms:
## Summary of Changes

- FilePathParam:
  - Internally uses std::filesystem now. This omits operating system dependencies for path format (but keeping interface).
  - Reworked FilePathParam flags -> added tests for flags
  - Added option to allow only specific file extensions for FilePathParam 
  - Global notifications and pop-ups: use weak_ptr for open flag and do not log output when opening pop-up to allow individual log level
  - Removed log mesage parsing for notification pop-ups -> moved notification pushes to FilePathParam 
- Removed useless and expensive UTF8 conversions (FileUtils, GUI, SDFFont)
- Fixed OverlayRenderer fbo usage
- Some minor UX fixes, code clean-up ...


core/include/mmcore/param/FilePathParam.h
```diff
@@ -30,70 +28,45 @@ namespace param {
...
 
         /**
          * Ctor.
          *
          * @param initVal The initial value
          * @param flags The flags for the parameter
+         * @param exts The required file extensions for the parameter
          */
-        FilePathParam(const wchar_t *initVal, UINT32 flags = FLAG_NONE);
+        FilePathParam(const std::string& initVal, Flags_t flags = Flag_File, const Extensions_t& exts = {});
```
moritz-h, line 57: Maybe an std::filesystem::path overload is also useful.

core/include/mmcore/param/FilePathParam.h
```diff
@@ -114,77 +87,75 @@ namespace param {
          * @param setDirty If 'true' the dirty flag of the owning parameter
          *                 slot is set and the update callback might be called.
          */
-        void SetValue(const vislib::StringA& v, bool setDirty = true);
+        void SetValue(const std::string& v, bool setDirty = true);
+        void SetValue(const vislib::TString& v, bool setDirty = true);
```
moritz-h, line 91: Maybe an std::filesystem::path overload is also useful.

reinago (approved):

PR #847, opened 2021-07-14
serialize graph: call argument fix

braunms:
Removed additional empty string argument for 'mmCreateCall' in graph serialization (e.g. used in screenshots).

Solves error "mmCreateCall:  expects 3 arguments of type (string, string, string) but has 4" on loading projects from png files.


geringsj:
This is already fixed in #805 (https://github.com/UniStuttgart-VISUS/megamol/pull/805/commits/0536630ed87bb26ff2f7d41876c738c62da4990c)

PR #859, opened 2021-07-29
Profiling tests for Hagen

reinago:
- added profiling capabilities to calls (activate via ENABLE_PROFILING in CMake)
- a bit of UX in configurator (nice graphs)
- fix for event transport across ContextToGL modules
- added CallCapabilities to enable asking Calls at runtime whether you should expect some kind of API at the other end


core/src/Call.cpp
```diff
@@ -55,20 +66,33 @@ bool Call::operator()(unsigned int func) {
...
+        const auto startTime = std::chrono::high_resolution_clock::now();
+        bool gl_started = false;
+        if (caps.OpenGLRequired()) {
+            gl_started = CallProfiling::qm->Start(this, this->callee->GetCoreInstance()->GetFrameID(), func);
+        }
 #endif
         res = this->callee->InCall(this->funcMap[func], *this);
+#ifdef PROFILING
+        if (gl_started) {
+            CallProfiling::qm->Stop(this->callee->GetCoreInstance()->GetFrameID());
```
geringsj, line 80: If we want to get rid of the core instance one day we need some other way to access the FrameID.

frontend/services/opengl_glfw/OpenGL_GLFW_Service.cpp
```diff
@@ -797,6 +801,10 @@ void OpenGL_GLFW_Service::postGraphRender() {
 
     ::glfwSwapBuffers(m_pimpl->glfwContextWindowPtr);
 
+#ifdef PROFILING
+    core::CallProfiling::CollectGPUPerformance();
```
geringsj: I would really like to avoid the GL Service knowing about calls and such. I think the `CollectGPUPerformance()` hints that we want a service to do the performance collection after the buffer swap?

core/src/MegaMolGraph.cpp
```diff
@@ -633,6 +633,14 @@ bool megamol::core::MegaMolGraph::add_call(CallInstantiationRequest_t const& req
         return false;
     }
 
+#ifdef PROFILING
+    std::vector<std::string> callbacks(call_description->FunctionCount());
+    for (uint32_t x = 0; x < call_description->FunctionCount(); ++x) {
+        callbacks[x] = call_description->FunctionName(x);
+    }
+    call->setProfilingInfo(callbacks);
+#endif
```
geringsj: I think we could move this to `CallAutoDescription<T>::CreateCall()` or decide to give each call all its function names regardless whether `PROFILING` is active or not?

geringsj (dismissed):
Looks good. Some clean up or refactoring could be done to remove profiling from `MegaMolGraph` and `OpenGL_GLFW_Service`, but the current implementation is ok for now.

frontend/services/profiling_service/Profiling_Service.hpp
```diff
@@ -0,0 +1,39 @@
+/*
+ * Screenshot_Service.hpp
```
geringsj, line 2: Might want to change that

geringsj (dismissed):
Nothing out of the ordinary

geringsj (dismissed):

geringsj (dismissed):

geringsj (approved):

PR #870, opened 2021-08-25
Assao

sellendk:
Contains Intels ASSAO implementation (currently without the highest quality setting, i.e. without the adaptive part) and a new module called ASSAO which can replace the current SSAO ScreenSpaceEffect module.

## Summary of Changes
Mapped Intels ASSAO Direct3D implementation to OpenGL. A new module has been added called "ASSAO" which takes a depth and a normal texture, and the camera as inputs and has various parameters to adjust the settings.

## References and Context
See https://github.com/GameTechDev/ASSAO/blob/master/Projects/ASSAO/ASSAO/ASSAODX11.cpp for the original implementation and https://software.intel.com/content/www/us/en/develop/articles/adaptive-screen-space-ambient-occlusion.html for more details.

## Test Instructions
- Take the CompositingExample.lua and replace the current SSAO (ScreenSpaceEffect) module with the new ASSAO module


invor:
Two more requests:
- remove Sampler.hpp and Texture2DView.hpp from compositing plugin and use glowl files instead
- add our copyright notice below the Intel copyright because we did modify the code quite a bit to use glsl
```
/**
 * MegaMol
 * Copyright (c) 2021, MegaMol Dev Team
 * All rights reserved.
 */
```

sellendk:
Done

invor (approved):

PR #872, opened 2021-09-08
Global Hotkeys and CUE support

reinago:
TODO:
- [x] actual CUE support
- [x] keyboard event routing through CommandRegistry
- [x] lua ExecuteCommand()
- [x] command naming from module side
- [x] cleanup of ButtonParam API
- [x] hotkey editor? like this https://github.com/CedricGuillemet/ImHotKey ? -> no, simpler, but good enough for now.
- [x] hotkey state needs to be saved together with GUI state I guess


core/include/mmcore/LuaInterpreter.h
```diff
@@ -392,9 +394,14 @@ bool megamol::core::LuaInterpreter<T>::RunString(
                         if (getDouble(1, r)) {
                             result = std::to_string(r);
                         } else {
-                            result = "Result is a complex type.";
-                            megamol::core::utility::log::Log::DefaultLog.WriteError("Lua execution returned complex type");
-                            good = false;
+                            bool b;
+                            if (getBool(1, b)) {
```
geringsj, line 398: `to_string(bool)` returns `"0"` or `"1"` and i wonder whether we want `result \in {"true", "false"}` instead.

core/include/mmcore/LuaInterpreter.h
```diff
@@ -392,9 +394,14 @@ bool megamol::core::LuaInterpreter<T>::RunString(
...
+                                result = std::to_string(b);
```
geringsj, line 399: `to_string(bool)` returns `"0"` or `"1"` and i wonder whether we want `result \in {"true", "false"}` instead.

frontend/services/gui/src/windows/HotkeyEditor.cpp
```diff
@@ -0,0 +1,313 @@
...
+                                                    case (megamol::frontend_resources::Command::parent_type_c::
+                                                            PARENT_GUI_WINDOW):
+                                                        cmd.effect = this->parent_gui_window_lambda;
+                                                        break;
+                                                    case (megamol::frontend_resources::Command::parent_type_c::
+                                                            PARENT_GUI_WINDOW_HOTKEY):
+                                                        cmd.effect = this->parent_gui_window_hotkey_lambda;
+                                                        break;
+                                                    }
+                                                    this->command_registry_ptr->add_command(cmd);
```
geringsj, line 273: I am not sure why the GUI Hotkey window needs to serialize and reload the whole command state. If Commands come from a certain graph/GUI state, wouldnt the Command Registry be repopulated with the same commands when loading the project file? Doesnt it suffice to just save the (maybe non-default) key<->command bindings instead of the whole command registry?

core/include/mmcore/MegaMolGraph.h
```diff
@@ -143,6 +144,13 @@ class MEGAMOLCORE_API MegaMolGraph { //: public serializable, public deferrable_
 
     MegaMolGraph_Convenience& Convenience();
 
+    frontend_resources::Command::EffectFunction Parameter_Lambda = [&](const frontend_resources::Command* self) {
+        auto my_p = this->FindParameter(self->parent);
+        if (my_p != nullptr) {
+            my_p->setDirty();
+        }
+    };
```
geringsj, line 152: I would be glad if we found a way to somewhow keep such code related to the commands out of the Graph.

core/src/MegaMolGraph.cpp
```diff
@@ -267,8 +529,23 @@ bool megamol::core::MegaMolGraph::add_module(ModuleInstantiationRequest_t const&
...
+            auto ps = dynamic_cast<param::ParamSlot*>((*child).get());
+            if (ps != nullptr) {
+                auto p = ps->Param<param::ButtonParam>();
+                if (p != nullptr) {
+                    frontend_resources::Command c;
+                    c.key = p->GetKeyCode();
+                    c.parent = ps->FullName();
+                    c.name = module_ptr->Name().PeekBuffer() + std::string("_") + ps->Name().PeekBuffer();
+                    c.effect = this->Parameter_Lambda;
+                    m_command_registry->add_command(c);
```
geringsj, line 544: I think we can abstract this intrusive code away when we get the Graph subscriber / Graph diffs / Graph serialization infrastructure up.

geringsj (dismissed):
Looks good overall. Regarding the JSON serialization i think it suffices to just save and reload the (key, command) pair, and maybe we can leave the serialization code to the Command Registry/Service itself. Because in the current state the Hotkey Window handles some pretty intimate Command Registry logic?

frontend/services/gui/src/windows/HotkeyEditor.cpp
reinago: we could add tracking of non-default keys. but the problem is: if modules are serialized in a different order than they were created in the original project file, other commands will get the overlapping hotkey (first come, first serve) - like when you have multiple view::resetView commands. then you need to have all the hotkey information anyway.

core/include/mmcore/LuaInterpreter.h
reinago: we probably need to think about this some more, string-to-boolean in lua seems to be a nontrivial thing. I did not know that

core/include/mmcore/MegaMolGraph.h
reinago: we can define this lambda anywhere else if you prefer that. even in the commandregistry itself...

moritz-h (approved):

PR #892, opened 2021-10-01
SMAA

sellendk:
Contains Subpixel Morphological AntiAliasing (SMAA). Also splits the "ScreenSpaceEffect" module into two separate modules "AntiAliasing" and "SSAO". Contains a new module "PNGDataSource"

## Summary of Changes
- Added SMAA to megamol
- Separated ScreenSpaceEffect module into 2 modules: one for AntiAliasing and one for SSAO
  - In the AntiAliasing module, you can now choose between SMAA and FXAA
  - In the SSAO module, you can choose between simple SSAO and ASSAO (see pull request #870)
- Added a new module "PNGDataSource" which just loads a .png image and sets it as [glowl::Texture2D](https://github.com/invor/glowl/blob/master/include/glowl/Texture2D.hpp) pointer to the output

## References and Context
- See the paper on SMAA [here](http://www.iryoku.com/smaa/)
- The original code can be found [here](https://github.com/iryoku/smaa)

## Test Instructions
Execute the CompositingExample.lua and replace the ScreenSpaceEffect module (with FXAA set) with the new AntiAliasing module and select SMAA


plugins/compositing_gl/src/AntiAliasing.cpp
```diff
@@ -0,0 +1,722 @@
...
+        m_smaa_neighborhood_blending_prgm = std::make_unique<GLSLComputeShader>();
+        m_copy_prgm = std::make_unique<GLSLComputeShader>();
+
+        vislib_gl::graphics::gl::ShaderSource fxaa_src_c;
+        vislib_gl::graphics::gl::ShaderSource smaa_edge_detection_src_c;
+        vislib_gl::graphics::gl::ShaderSource smaa_blending_weights_src_c;
+        vislib_gl::graphics::gl::ShaderSource smaa_neighborhood_blending_src_c;
+        vislib_gl::graphics::gl::ShaderSource copy_src_c;
+
+        auto ssf = std::make_shared<megamol::core_gl::utility::ShaderSourceFactory>(
```
moritz-h: It would be better to use the new shader factory for a completely new module.

plugins/compositing_gl/shaders/Compositing.btf
```diff
@@ -34,9 +39,34 @@
 
     <shader name="fxaa">
         <snippet type="version">430</snippet>
-        <snippet name="Main"             type="file">comp/fxaa_c.glsl</snippet>
+        <snippet name="Main"             type="file">comp/AntiAliasing/fxaa_c.glsl</snippet>
     </shader>
 
+    <namespace name="smaa">
```
moritz-h: It would be better to use the new shader factory for a completely new module, without any new BTF stuff. BTF is deprecared and should die anyway :-)

plugins/compositing_gl/src/SMAASearchTex.h
```diff
@@ -0,0 +1,1070 @@
+/**
```
moritz-h: Would be nice to not just place thirdparty code right within the megamol code.
Best solution would be if this code can be a an fully external. If that is not possible, the second best option would be if this would be an inlined external library such as glad [see here](https://github.com/UniStuttgart-VISUS/megamol/tree/master/externals/glad). And if that is also not possible, please put it at least in a `plugins/compositing_gl/3rd` directory, such as the datatools and infovis plugin already are doing. A short readme with the source of this 3rd party code would be also nice to have there.

core/include/mmcore/profiler/TimeMeasure.h
```diff
@@ -0,0 +1,34 @@
+/*
```
moritz-h: Please use the new default copyright text, copy from here with the correct year: https://github.com/UniStuttgart-VISUS/megamol/blob/master/plugins/doc_megamol101_gl/src/megamol101_gl.cpp

core/include/mmcore/profiler/TimeMeasure.h
```diff
@@ -0,0 +1,34 @@
...
+ * TimeMeasure.h
+ *
+ * Copyright (C) 2021 by Universitaet Stuttgart (VISUS).
+ * All rights reserved.
+ */
+
+#ifndef TIME_MEASURE_H_INCLUDED
```
moritz-h: replace by `#pragma once`

plugins/compositing_gl/src/AntiAliasing.h
```diff
@@ -0,0 +1,310 @@
+/*
+ * AntiAliasing.h
+ *
+ * Copyright (C) 2021 by Universitaet Stuttgart (VISUS).
+ * All rights reserved.
+ */
+
+#ifndef ANTI_ALIASING_H_INCLUDED
```
moritz-h: `#pragma once`

plugins/compositing_gl/src/PNGDataSource.h
```diff
@@ -0,0 +1,68 @@
+/*
+ * PNGDataSource.h
+ *
+ * Copyright (C) 2015-2015 by CGV (TU Dresden)
+ * Alle Rechte vorbehalten.
+ */
+
+#ifndef MEGAMOL_DATATOOLS_PNGDATASOURCE_H_INCLUDED
```
moritz-h: only `#pragma once`

plugins/compositing_gl/src/PNGDataSource.h
```diff
@@ -0,0 +1,68 @@
...
+
+#include "mmcore/Call.h"
+#include "mmcore/CalleeSlot.h"
+#include "mmcore/Module.h"
+#include "mmcore/param/ParamSlot.h"
+
+#include "glowl/Texture2D.hpp"
+
+namespace megamol {
+namespace compositing {
```
moritz-h: should be compositing_gl namespace

plugins/compositing_gl/src/AntiAliasing.h
```diff
@@ -0,0 +1,310 @@
...
+#include "vislib_gl/graphics/gl/GLSLComputeShader.h"
+
+#include "glowl/BufferObject.hpp"
+#include "glowl/Texture2D.hpp"
+
+#include <glm/ext.hpp>
+#include <glm/glm.hpp>
+
+namespace megamol {
+namespace compositing {
```
moritz-h: should be `compositing_gl` namespace. Can use C++17 style.

moritz-h (changes requested):

plugins/compositing_gl/src/SMAASearchTex.h
invor: Inlined external library or placed in a dedicated `3rd` folder I would supprt and prefer as well. I don't think we want to clone the complete external repository just to use two header files.

plugins/compositing_gl/src/AntiAliasing.cpp
sellendk: Done.

plugins/compositing_gl/shaders/Compositing.btf
sellendk: Done.

plugins/compositing_gl/src/SMAASearchTex.h
sellendk: Added a 3rd party folder. Done.

core/include/mmcore/profiler/TimeMeasure.h
sellendk: Done.

core/include/mmcore/profiler/TimeMeasure.h
sellendk: Done.

plugins/compositing_gl/src/AntiAliasing.h
sellendk: Done.

plugins/compositing_gl/src/PNGDataSource.h
sellendk: Done.

plugins/compositing_gl/src/PNGDataSource.h
sellendk: Done.

plugins/compositing_gl/src/AntiAliasing.h
sellendk: Done.

core/src/profiler/TimeMeasure.cpp
```diff
@@ -1,7 +1,6 @@
-/*
- * TimeMeasure.cpp
- *
- * Copyright (C) 2021 by Universitaet Stuttgart (VISUS).
+/**
+ * MegaMol
+ * Copyright (c) 2022, MegaMol Dev Team
```
moritz-h, line 3: Please do never update any Copyright years and use always the original one. Reason: https://stackoverflow.com/a/2391555
(Exception would be if there is given a range 2000-2021, then the range should be updated. But a range is generally not required and we can omit this, to avoid needing to update it.)

moritz-h (changes requested):

core/include/mmcore/profiler/TimeMeasure.h
```diff
@@ -0,0 +1,30 @@
+/**
+ * MegaMol
+ * Copyright (c) 2022, MegaMol Dev Team
```
moritz-h: Please do never update any Copyright years and use always the original one. Reason: https://stackoverflow.com/a/2391555
(Exception would be if there is given a range 2000-2021, then the range should be updated. But a range is generally not required and we can omit this, to avoid needing to update it.)

moritz-h (dismissed):

core/include/mmcore/profiler/TimeMeasure.h
sellendk: Ah, sorry, did not know about this and thanks for clarifying :)

invor (approved):

PR #922, opened 2021-11-05
Separate FBO for GUI Service

braunms:
Added separate FBO to GUI service and replaced image presentation hack for the GUI. 

## Summary of Changes
- The GUI service is now added as additional entry point providing complete GUI rendering in result image utilized by image presentation service.
- Replaced fbo blitting with classic texture drawing to support blending of multiple result images.

### DIFF
--- a/frontend/services/image_presentation/ImagePresentation_Service.cpp
+++ b/frontend/services/image_presentation/ImagePresentation_Service.cpp
@@ -15,11 +15,9 @@
 
 #include "ImageWrapper_to_GLTexture.h"
 #include "ImagePresentation_Sinks.hpp"
-#include "GUI_Service.hpp"
 
 #include "LuaCallbacksCollection.h"
-
-#include "mmcore/view/AbstractView_EventConsumption.h"
+#include "RenderInput.h"
 
 #include <any>
 #include <utility>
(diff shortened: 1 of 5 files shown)


frontend/services/image_presentation/ImagePresentation_Service.cpp
```diff
@@ -15,6 +15,7 @@
 
 #include "ImageWrapper_to_GLTexture.h"
 #include "ImagePresentation_Sinks.hpp"
+#include "GUI_Service.hpp"
```
geringsj: To avoid including the whole GUI Service i would extract the GUI entry point callbacks into a separate header, which both GUI Service and ImagePresentation Service can include. The actual implementation of the callbacks can stay in the GUI Service cpp.

frontend/services/image_presentation/ImagePresentation_Service.cpp
```diff
@@ -155,11 +156,13 @@ std::tuple<
...
-                std::function{megamol::core::view::view_rendering_execution},
+            return EntryPointInitFunctions{std::function{megamol::core::view::view_rendering_execution},
                 std::function{megamol::core::view::view_init_rendering_state},
-                std::function{megamol::core::view::get_gl_view_runtime_resources_requests}
-            };
+                std::function{megamol::core::view::get_gl_view_runtime_resources_requests}};
+        } else if (auto gui_ptr = static_cast<megamol::frontend::GUI_Service*>(ptr); gui_ptr != nullptr) {
+            return EntryPointInitFunctions{std::function{megamol::frontend::GUI_Service::gui_rendering_execution},
+                std::function{megamol::frontend::GUI_Service::gui_init_rendering_state},
+                std::function{megamol::frontend::GUI_Service::get_gl_gui_runtime_resources_requests}};
```
geringsj: Does the `else if` get triffered only if the `void* ptr` is a `megamol::core::Module*`?

frontend/services/image_presentation/ImagePresentation_Sinks.hpp
```diff
@@ -6,12 +6,14 @@
...
 
 #include "ImageWrapper.h"
+#include <glowl/GLSLProgram.hpp>
 
 namespace megamol::frontend {
 
     struct glfw_window_blit {
         unsigned int gl_fbo_handle = 0;
         unsigned int fbo_width = 0, fbo_height = 0;
+        std::shared_ptr<glowl::GLSLProgram> blit_shader;
```
geringsj: To avoid pulling GL headers into the Image Presentation Service i would try to hide the glowl headers and shader class. Maybe via pimpl or so?

PR #926, opened 2021-11-09
Profiling features

reinago:
generalize profiling:
- enable logging
- preserve fast live info
- add code regions for modules
- nuke leftovers from previous approach
- bonus: interactive log console


geringsj:
As the profiling grows it may be time to merge the Frame Statistics Service into it. It seems they are doing similar things.

braunms (approved):

PR #927, opened 2021-11-10
GUI Service: Gui As Entry Point + CPU Backend Preparation

geringsj:
Imgui renderins into own FBO as Image Presentation Entry Point. Prepares GUI Service to use other Imgui backends than OpenGL.
Adds rendering result blending via shader to GLFW Window Sink.

## TODOs
* refactor `glfw_window_blit` to only use GLSLProgram shader and restore previous GL state


braunms (dismissed):
LGTM!

tobiasrau (approved):

PR #936, opened 2021-11-15
removed shader factory and gl module registration out of coreinstance

tobiasrau:
## Summary of Changes

This removes GL dependencies in the non-GL version of the core.


CMakeLists.txt
```diff
@@ -84,6 +84,10 @@ endif()
 # Plugins
 add_subdirectory(plugins)
 if (BUILD_FRONTEND)
+  if (ENABLE_GL)
+    target_link_libraries(plugins INTERFACE core_gl)
+    target_sources(plugins INTERFACE $<TARGET_OBJECTS:core_gl>)
```
moritz-h, line 89: This feels conceptionally wrong. the core should not be a public depedency of the plugins. Individual plugins already links the core_gl (private). Also putting core objects into the plugins target is probably not the best place.

core_gl/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ if(BUILD_CORE)
   file(GLOB_RECURSE shader_files RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}" "Shaders/*")
 
   # Target definition
-  add_library(${PROJECT_NAME} STATIC ${public_header_files} ${header_files} ${shader_files} ${source_files})
+  add_library(${PROJECT_NAME} OBJECT ${public_header_files} ${header_files} ${shader_files} ${source_files})
```
moritz-h, line 25: When making the core_gl an object library, probably the core_gl objects can be attached to the core lib itself when gl is enabled.

plugins/compositing_gl/src/ScreenSpaceEffect.cpp
```diff
@@ -15,19 +15,20 @@
...
+                                                             , m_output_texture_hash(0)
+                                                             , m_mode("Mode", "Sets screen space effect mode, e.g. ssao, fxaa...")
+                                                             , m_ssao_radius("SSAO Radius", "Sets radius for SSAO")
+                                                             , m_ssao_sample_cnt("SSAO Samples", "Sets the number of samples used SSAO")
+                                                             , m_output_tex_slot("OutputTexture", "Gives access to resulting output texture")
+                                                             , m_input_tex_slot("InputTexture", "Connects an optional input texture")
+                                                             , m_normals_tex_slot("NormalTexture", "Connects the normals render target texture")
+                                                             , m_depth_tex_slot("DepthTexture", "Connects the depth render target texture")
+                                                             , m_camera_slot("Camera", "Connects a (copy of) camera state") {
```
moritz-h: This looks like a clang-format problem. There should not be that much spaces.

moritz-h (changes requested):

moritz-h (approved):

PR #954, opened 2021-11-29
software render backend for imgui rendering

braunms:
Applied software render backend proposed in #942 for alternative GUI rendering.

## Summary of Changes
- Moved GUI render backend functionality to separate class.
- Allow alternative software GUI rendering.
- Fix: Broken shader pointer of picking
- Fix: Clear fbo is gui is set invisible
- Fix: increase priority of command service to process only keyboard events not consumed by gui

## Test Instructions
- Configure CMake: ENABLE_GL=OFF 
- Execute MegaMol with project file only containing: 

```
mmRenderNextFrame()
mmRenderNextFrame()
mmScreenshotEntryPoint("GUI_Service", "cpu_screenshot.png")
mmQuit()
```


core/include/mmcore/utility/Picking.h
```diff
@@ -112,12 +114,13 @@ class PickingBuffer {
...
     std::map<unsigned int, std::vector<Interaction>> available_interactions;
     ManipVector_t pending_manipulations;
+    bool enabled;
+
 #ifdef WITH_GL
     int prev_fbo = 0;
     glowl::FramebufferObject* fbo = nullptr;
-    glowl::GLSLProgram* fbo_shader = nullptr;
+    glowl::GLSLProgram* shader = nullptr;
 #endif
```
geringsj: Why not drop the * ? Use plain objects.

core_gl/src/utility/Picking_gl.cpp
```diff
@@ -10,23 +10,28 @@
...
     }
 
+
 PickingBuffer::~PickingBuffer() {
 
-    delete this->fbo;
+    if (this->shader != nullptr)
+        delete this->shader;
+    if (this->fbo != nullptr)
+        delete this->fbo;
```
geringsj: Use managed pointers or objects directly.

core/include/mmcore/utility/Picking.h
braunms: There wont't be any OpenGL realted headers be available in the core even when WITH_GL is defined. So, the only option here is to use forward declared classes via raw pointers in the header.

core_gl/src/utility/RenderUtils.cpp
```diff
@@ -102,6 +96,41 @@ bool RenderUtils::CreateShader(std::shared_ptr<glowl::GLSLProgram>& out_shader_p
 }
 
 
+bool RenderUtils::CreateShader(
+    glowl::GLSLProgram** out_shader_ptr, const std::string& vertex_code, const std::string& fragment_code) {
```
geringsj: Would a `glowl::GLSLProgram &` work here too?

frontend/services/gui/src/GUI_Service.cpp
```diff
@@ -387,12 +377,12 @@ bool GUI_Service::gui_rendering_execution(void* void_ptr,
...
-    gui_ptr->GetFBODataGL(fbo_color_buffer_gl_handle, fbo_width, fbo_height);
+    gui_ptr->GetFBOData_GL(fbo_color_buffer_gl_handle, fbo_width, fbo_height);
 
     result_image =
         megamol::frontend_resources::wrap_image({fbo_width, fbo_height}, fbo_color_buffer_gl_handle, channels);
 #else
-    auto fbo = gui_ptr->getFBOHandle();
+    auto fbo = gui_ptr->GetFBOData_CPU();
     result_image =
         megamol::frontend_resources::wrap_image({fbo->getWidth(), fbo->getHeight()}, fbo->colorBuffer, channels);
```
geringsj: If GetFBOData() would either return an `ImageWrapper` the distinction between `_GL` and `_CPU` could be removed. Or instead of `fbo_color_buffer_gl_handle/fbo->colorBuffer` one could return a generic type that can be wrapped by the ImageWrapper.

braunms: Good idea, see commit 2b4ed5d

core/include/mmcore/utility/Picking.h
braunms: Edit: Managed pointers also allow incomplete types, see commit 0e88026

core_gl/src/utility/Picking_gl.cpp
braunms: See commit 0e88026

core_gl/src/utility/RenderUtils.cpp
braunms: Obsolete due to commit 0e88026

geringsj:
I suggest to activate the GUI Service in no-GL mode for further debugging, i.e. looking at GUI screenshots done with `mmScreenshotEntryPoint("GUI Service", "gui_screenshot.png")`. The message by the Image Presentation that GLFW is missing is just a warning and has no effect on GL/no-GL functionality.

```
diff --git a/frontend/main/src/main.cpp b/frontend/main/src/main.cpp
index c30606b2c..5bb5a9133 100644
--- a/frontend/main/src/main.cpp
+++ b/frontend/main/src/main.cpp
@@ -180,8 +180,8 @@ int main(const int argc, const char** argv) {
     megamol::frontend::FrontendServiceCollection services;
     if (with_gl) {
         services.add(gl_service, &openglConfig);
-        services.add(gui_service, &guiConfig);
     }
+    services.add(gui_service, &guiConfig);
     services.add(lua_service_wrapper, &luaConfig);
     services.add(screenshot_service, &screenshotConfig);
     services.add(framestatistics_service, &framestatisticsConfig);

```

geringsj (approved):

PR #964, opened 2021-12-14
Fixes to project loading and LogConsole

reinago:
depends on #926, no way around it

- would allow projects themselves to ask for their path
- makes the camera less confused when using mmRenderNextFrame inside a project
- prevents explosion on next frame with new LogConsole features


reinago:
changes are already in master

PR #975, opened 2022-01-21
thinking about the data source, background loading and data transport mess

reinago:
open questions:


plugins/datatools/include/datatools/io/dataformat/DataFormat.h
```diff
@@ -0,0 +1,152 @@
...
+    // TODO: Adrian's LRUCache here since we need most of the functionality anyway, so why not.
+    using FrameCollection = std::unordered_map<FrameIndexType, FrameType>;
+
+    // TODO generic EnumerateFrames that only files the frame indices with empty frames for now?
+
+    std::unique_ptr<FrameType> ReadFrame(FrameIndexType idx) {
+        // here we should actually grab the frames from some background thread that reads ahead and stuff?
+        FrameType f;
+        f.Read(IndexToIStream(idx));
+        return std::make_unique<FrameType>(f);
```
geringsj: Use `std::move(f)` and demand that FrameType is moveable (C++ concepts?)

plugins/datatools/include/datatools/io/dataformat/ImageCalls.h
```diff
@@ -0,0 +1,275 @@
...
+        ASSERT(width * height * depth * channelType.ByteSize() == data.size());
+        this->data = data;
+        this->width = width;
+        this->height = height;
+        this->depth = depth;
+        this->elementType = channelType;
+    }
+
+    // data will be controlled by this instance, caller loses access!
+    void SetData(uint8_t* data, std::size_t count, ImageElementType channelType, SizeType width = 1, SizeType height = 1, SizeType depth = 1) {
```
gralkapk: That's a unique_ptr then

plugins/datatools/include/datatools/io/dataformat/ImageCalls.h
```diff
@@ -0,0 +1,275 @@
...
+        SetData(vec, channelType, width, height, depth);
+    }
+
+    template <typename T>
+    constexpr T* ViewAs() const {
+        return static_cast<const T*>(data.data());
+    }
+
+    template <typename ResultType>
+    std::vector<ResultType> GetCopy() {
```
gralkapk: The switch-case part in the following functions can be in a separate method using a functor for copy/set/get functionality

plugins/datatools/include/datatools/io/dataformat/ImageCalls.h
```diff
@@ -0,0 +1,275 @@
...
+        case ImageElementType::FLOAT:
+            CopyInto<ResultType, float>(out);
+            break;
+        case ImageElementType::DOUBLE:
+            CopyInto<ResultType, double>(out);
+            break;
+        default:
+            throw std::logic_error("ImageFrame::GetCopy: invalid elementType");
+        }
+        return std::move(out);
```
gralkapk: No move on return

plugins/datatools/include/datatools/io/dataformat/ImageCalls.h
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    template<typename Dest, typename Source>
+    void SetAbsolute(SizeType index, Source val) {
+        AccessAs<Dest>()[index] = static_cast<Dest>(val);
+    }
+
+    template<typename Dest, typename Source>
+    void CopyInto(std::vector<Dest>& out) {
+        std::transform(ViewAs<Source>(), ViewAs<Source>() + NumElements(), std::back_inserter(out),
```
gralkapk, line 269: Method hides whether reserve or resize should be used for out

plugins/datatools/include/datatools/io/dataformat/ImageCalls.h
```diff
@@ -0,0 +1,275 @@
...
+// which variants do we want to handle dynamically? different channel types probably. dimensionality is unlikely.
+// different numbers of components?
+
+// TODO: get rid of NumComponents. There is only one, keep multiple images if you want more channels.
+// exception: RGBA8 as UINT32 ? Does it matter?
+
+// typename ChannelType does not work, that is not a static decision.
+// For some data sources you only know after actually opening the file what's in it.
+// TODO: volume ist ein byte-vektor und wir benutzen span. fertig.
+template<uint8_t Dimensions>
```
gralkapk, line 21: Do we need dimension as static information since we do not change available methods?

reinago: I was wondering whether we should disable some of the methods depending on Dimensions. I am really not sure whether we should just treat dims dynamically too. A transfer function is a perfectly OK volume, strictly speaking. It's a bit thin, but not broken or something...

gralkapk: With the dimension discerning types, we have either three statically different call transporting 1D, 2D, or 3D volumes, or yet another dispatch. I would argue for dimension as template parameter, if we have a significant gain from it.

PR #978, opened 2022-01-27
MegaMol Project Concept

geringsj:
Uses current lua script file path to determine a project directory, allowing FilePath Parameters to load files relative to that directory. The project directory is injected into file parameters of newly created graph modules.

## TODO
- [X] Should File Path Parameters automatically check for files beeing inside the project dir, and use relative paths for those?
* What happens when saving and loading the project again?
   - [X] MegaMolGraph Serialization preserves relative file path (uses `param->ValueString()`)
   - [x] GUI Serialization preserves relative file path 
* ~~What happens when stacking project files?~~ FilePathParams remember their respective project file
* ~~Optional Lua Callback Arguments (optional project name argument)~~
* ~~Do we really want to clear() the graph?~~
- ~~`vislib::sys::File` and `vislib::sys::File::Exists()` must die~~ Only FilePathParam knows about project, re-routing all file loading through project dir too intrusive
- ~~Screenshots and other outputs go to project dir automatically?~~ User configures screenshot output 
- [ ] Unify UX: Configurator project open, project add, file load, file drop
- [ ] Adjust example projects for new file path structure

## Test Instructions
The implementation works for all code paths going though the Project Loader Service, e.g. Drag & Drop of files into the window and "Load Project" in the GUI.
For example, modify the `infovis.lua` example to access the example data set relative to the examples directory.
```lua
-- snip --

-- old: mmSetParamValue("::CSVDataSource_1::filename",[=[../examples/sampledata/iris.csv]=])
mmSetParamValue("::CSVDataSource_1::filename",[=[sampledata/iris.csv]=])
```

grafik

### DIFF
--- a/core/include/mmcore/param/FilePathParam.h
+++ b/core/include/mmcore/param/FilePathParam.h
@@ -118,7 +118,8 @@ public:
      *
      * @return Return 0 for success, flags with failed check otherwise.
      */
-    static Flags_t ValidatePath(const std::filesystem::path& p, const Extensions_t&, Flags_t f);
+    static Flags_t ValidatePath(
+        const std::filesystem::path& p, const Extensions_t&, Flags_t f, const std::filesystem::path& project_dir = "");
 
 
     /**
--- a/core/include/mmcore/param/FilePathParam.h
+++ b/core/include/mmcore/param/FilePathParam.h
@@ -83,7 +83,7 @@ public:
      * @return The value of the parameter
      */
     std::filesystem::path Value() const {
-        return this->value;
+        return this->GetAbsolutePathValue(this->value);
     }
 
     /**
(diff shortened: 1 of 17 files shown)


core/include/mmcore/param/FilePathParam.h
```diff
@@ -132,18 +132,37 @@ class FilePathParam : public AbstractParam {
...
+    Extensions_t extensions;
 
     /** The file or directory path */
     std::filesystem::path value;
+
+    /**
+     * Absolute path to project directory. If empty, no project path available.
+     * This path is relative per guarantee of the frontend
+     */
+    std::filesystem::path project_directory;
```
geringsj, line 177: This might also be just a "prefix_path"

PR #980, opened 2022-02-07
Update manual.md with instructions for building on windows without visual studio

ebosseck:
Added instructions for building using VS Build Tools, so no Microsoft account is required for building under windows (VisualStudio Community Edition requires a Microsoft account after an initial trial period of 30 days),

## Summary of Changes

Added instructions for building using VS Build Tools, so no Microsoft account is required for building under windows

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -8,6 +8,7 @@
 - [Installation and Setup](#installation-and-setup)
     - [Building from Source](#building-from-source)
         - [Microsoft Windows](#microsoft-windows)
+        - [Microsoft Windows (Command Line)](#microsoft-windows-command-line)
         - [Linux (Ubuntu)](#linux-ubuntu)
     - [Command Line Arguments](#command-line-arguments)        
     - [Configuration File](#configuration-file)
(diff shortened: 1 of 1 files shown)


reinago:
Hi ebosseck, thank you for taking the time and updating the documentation!

A command-line build is quite useful also on windows, however I was wondering about some details:
- Ninja would be the better build tool (also inside VS - but that is another story). I am unsure whether the standalone VS build tools also come with ninja (VS does when you install CMake support)
- The configuration, generation, and build itself can be formulated much more concisely (see also the CI configuration), basically
```
cmake -G "Visual Studio 17 2022" -B build
cmake --build build --config Release
cmake --install build --config Release
```

To be honest, wouldn't it be much more interesting to have instructions for a development environment based on VS Code, either using VS build tools or - even better - ninja and clang. Would you be willing to look into that instead?

ebosseck:
Hi reinago,

From what I understood, the BuildTools are basically all the compiler parts of VisualStudio stripped down to the things necessary for compiling but without anything else. Honestly I'm not really an expert on C++ building systems, I'm just a HiWi for Sergej and I preferred to have a way to build megamol without requiring me to sign into an Microsoft account, so I attempted to stay as close to the original building instructions as possible. (My actual tasks require me to have megamol running, but at least for the Tasks I currently know of I do not have to write any code for megamol)

In case I should look into other build systems, please talk with Sergej, so that I get this officially as a task to work on, but I think there might be other people more qualified than me for looking into alternative build tools.

As IDE I'll probably look in the directions of CLion (The C++ IDE from Jetbrains/IntelliJ) or CodeBlocks if I have to, since I have them already installed. (VS Code might also be an option, but currently I dont see the benefits of VSCode over CodeBlocks (and CLion at least for Students), but IDEs are always also a matter of personal preference, apart from technical details. But I have not had a closer look into C++ IDEs jet, apart from the experience that VS Community requires a ms account after 30 days).

Greetings, eric

geringsj:
Hi Eric,

if you dont mind, would you try reinago's way to build/install in CMake and update your PR accordingly? If you use CLion we are also happy if you update the PR to help with setting up MegaMol in CLion - i guess they also support CMake and/or Ninja.

Regards,
Sergej

moritz-h:
Hi, I have already made a CLion manual for myself locally. You can use it if you want:

```markdown
# Build, Run and Debug MegaMol in CLion

- File => Settings => Build, Execution, Deployment => CMake
  - Add Release Profile
  - Add `-DCMAKE_INSTALL_PREFIX=../cmake-build-mm-install` to CMake options (for all profiles).
    - Path is relative to generation path (`cmake-build-*` dir). Choose path outside of generation path, so we can use common run configuration where we need to specify an absolute path.
  - (optional) Set Build options to more make jobs `-- -j 12`
- Open target selection box at top right of CLion window, next to run button => Edit Configurations
- '+' Button top left, add CMake Application
- Name `BUILD_INSTALL_ALL` (or whatever)
- Target: "All targets" (scroll up in the targets list, UI is tricky here)
- Executable: select any dummy binary or target for first build (will be changed later)
- "Before lunch" config, append "Install" after "Build"
- Click Ok, Run once (that files are installed)
- Go back to "Run Configurations"
- Executable: cmake-mm-install/bin/megamol (cannot use megamol.sh wrapper for debugging)
- Program arguments: `/path/to/project.lua`
- Environment variables: `LD_LIBRARY_PATH=../lib`
```

PR #1005, opened 2022-03-22
various fixes ...

braunms:
- fixed font size for cinematic stuff (font size 0 lead to infinite loop when calling mmSetGUIVisible(false) in project file)
- improved UX for scaling and font menu
- fixed some stuff in OverlayRenderer
- fixed bugs in logconsole auto completion
- added File_Any flag to FilePathParam to allow selection of files and directories in the same parameter
- ModernTrisoupRenderer: fixed mesh color, reset gl state of polygon mode
- Fixed GUI visibility toggling


frontend/services/gui/src/GUIManager.cpp
```diff
@@ -1149,18 +1149,23 @@ void GUIManager::draw_menu() {
             float scale = megamol::gui::gui_scaling.Get();
             if (ImGui::RadioButton("100%", (scale == 1.0f))) {
                 this->SetScale(1.0f);
+                ImGui::CloseCurrentPopup();
```
reinago: I am not really sure I want that. Good question actually.

braunms: Updated slightly in following commit to fit UX of other menus.

reinago:
I just looked at this and using the cinematic example I opened the log console, used mmHelp (using autocomplete) and could not scroll to the bottom anymore...

moritz-h (dismissed):

moritz-h (approved):

PR #1028, opened 2022-05-06
Ellipsoid

sellendk:
Adapted the moldyn_gl plugin to the new shader factory and added some more glyphs to the GlyphRenderer.

## Summary of Changes
- Change from old to new shader factory
- Added Superquadric to GlyphRenderer
- Pending: Gizmo-glyphs for GlyphRenderer
- Pending: Complete adapting to new shader factory and fix some bugs

## Problems
When adapting to the new shader factory, I encountered a problem regarding extensions. For example, in the fragment_extensions.inc.glsl (used in the Simple/Default Mode of the SphereRenderer - simple_sphere.frag.glsl):

```
#extension GL_ARB_explicit_attrib_location : enable   // glsl version 130
#extension GL_ARB_conservative_depth       : enable   // glsl version 130
layout (depth_greater) out float gl_FragDepth;
```

https://github.com/sellendk/megamol/blob/b875106e3d621c2631190a06f1568c51cf24d4d8/plugins/moldyn_gl/shaders/sphere_renderer/inc/fragment_extensions.inc.glsl#L2-L4

If I set the GL_ARB_conservative_depth extension from 'enable' to 'require' (which was the default state in the old shader factory where everything works fine), I get 2 error messages:

1.) ERROR: ..\ellipsoid\build\install\share\shaders\sphere_renderer\inc\fragment_extensions.inc.glsl:3: '#extension' : extension not supported: GL_ARB_conservative_depth

2.) ERROR: ..\ellipsoid\build\install\share\shaders\sphere_renderer\sphere_simple.frag.glsl:4: '#extension' : extra tokens -- expected newline

When setting to 'enable' again, everything works as intended.

## Test Instructions
To re-create the problem, just change 'enable' to 'require' in the mentioned file above and start testspheres.lua - no further changes needed.


straubar:
Some extensions don't seem to be supported by GLSLang. You can have a look here:
https://github.com/KhronosGroup/glslang/blob/master/glslang/MachineIndependent/Versions.h

If it is not supported, require will then cause an error. Enable, on the other hand, results only in a warning (that is ignored in MegaMol).

moritz-h:
We don't really care about full glslang validation and only using it for preprocessing the shader includes. With #1035 glslang should ignore these kind of errors and it will be possible to "require" any extension.

moritz-h (approved):

Issue #1038, opened 2022-05-20
File selection not possible for special setup of the FilePathParam


schatzkn:
When using the "param::FilePathParam::FilePathFlags_::Flag_NoExistenceCheck" flag on creating a FilePathParam, proper selection of a file or folder via GUI is impossible. 

1. Create a module or modify a module to have a FilePathParam.
2. On creation of the FilePathParam set the flags (the second parameter of the constructor) to "param::FilePathParam::FilePathFlags_::Flag_NoExistenceCheck".
3. Upon running MegaMol, try to modify the file path of the created FilePathParam by navigating through the MegaMol-internal file system representation.
4. It should not be possible to select any folder or file via mouse interaction. Instead, the current folder is set to "." and nothing else happens.

**Expected behavior**
It should be possible to select files or folders normally without the unexpected jumping.

**Environment**
 - OS: Windows 11
 - GPU: Nvidia RTX 2080Ti
 - Driver version 511.65

braunms:
It is mandatory to always state whether the FilePathParam should check for a file, directory, or both. You should use the flag Flag_File_ToBeCreated if you want to use the FilePathParam to hold a not existing file, e.g.

Of course, the FilePathParam flag interface should not provide flags that can not be used on their own. I am sorry, this will be fixed in a pull request soon...

PR #1039, opened 2022-05-20
FilePathParam Flags

braunms:
Separate internal flags to clarify usage of file flags - See #1038


core/include/mmcore/param/FilePathParam.h
```diff
@@ -27,18 +27,20 @@ namespace param {
  */
 class FilePathParam : public AbstractParam {
 public:
+    enum FilePathFlagsInternal_ : uint32_t { Internal_NoExistenceCheck = 1 << 2, Internal_RestrictExtension = 1 << 3 };
```
moritz-h: Shouldn't we just keep them in the original FilePathFlags_ enum?
The splitup would only make sense if we could make this separate enum private, but these are still required by some other classes, so they are probably required to stay public? Further as we are only using an enum here not enum class, the public interface of the FilePathParam does not change by splitting them, users still can just do `FilePathParam::Internal_NoExistenceCheck`. Also for refactoring in future it may is unexpected to have definition of bit sets separated over multiple enums, which could then lead to double definition of bits if someone is not careful.

moritz-h (changes requested):

core/include/mmcore/param/FilePathParam.h
braunms: Good point. See commit 0b0b1cd ...

moritz-h (approved):

PR #1042, opened 2022-05-20
Mostly stuff with triangle meshes

straubar:
Changes that are mostly related to triangle meshes, their handling, and rendering.
Additionally, there are some global fixes and improvements.

## Summary of Changes
- Added option for CMake to turn on and off CGAL, which is required for one of the new modules (turning it off, sets the module's availability to false)
- Added plane renderer to (transparently) render a (clip) plane, confined to the bounding box of the chained input
- Added (convenience) module for rendering 2D triangle meshes using RenderMDIMesh
- Added (convenience) module for rendering 3D triangle meshes using RenderMDIMesh
- Added module for providing triangle meshes for above renderers, inheriting from the STL reader from datatools
- Added module for writing STL meshes to file
- Added module for simplifying STL meshes, thereby reducing their complexity (requires CGAL)
- Improved datatools STL reader by reducing code complexity
- Improved texture combining, by adding parameters for respective weights
- Improved handling of bounding and clip boxes when combining them in the render chain
- Removed pre-compiled headers from mesh and mesh_gl


invor (commented):
- Connecting an new STLDataSource (i,e. with no file loaded) crashes MegaMol. Should be fixable.
- How come the STLWriter is located in the mesh plugin, but parts of the data source are located in datatools_gl?

straubar:
> * Connecting an new STLDataSource (i,e. with no file loaded) crashes MegaMol. Should be fixable.

The issue was fixed in the TriangleRenderer3D module, which could not handle empty input.

> * How come the STLWriter is located in the mesh plugin, but parts of the data source are located in datatools_gl?

There are two STLDataSources, one that is located in the datatools_gl and provides its data for CallTriMeshDataGL. The other data source in mesh_gl builds upon the reader, but provides its output for the TriangleMeshCall, and is itself called MeshSTLDataSource. The STLWriter is only provided for TriangleMeshCall, and thus also in mesh_gl.

Gegell:
> The issue was fixed in the TriangleRenderer3D module, which could not handle empty input.

This fix might not have captured all of the edge cases.
I ran into a similar (probably even the same) issue, where upon providing the Renderer with a `TriangleMeshCall` whose `indices` are pointed to a vector which contains no elements, MegaMol would crash. 

I believe this happens within the `TriangleMeshRenderer3D::getDataCallback` function due to the call to the `GPUMeshCollection::addMesh` function. I've at least tracked down the issue to there (after line 193 in `GPUMeshCollection.h` it crashes). Explicitly the error is: 
> Expression: can't dereference out of range vector iterator

straubar:
Added checks to enforce matching numbers of vertices, normals, and data values.
It is further intended that indices are always provided, even if they could be generated trivially. This is then the task of the data-providing module.

invor (approved):

PR #1054, opened 2022-06-22
Upgrade t-SNE, add UMAP

schulzch:
- Upgrade t-SNE (bugfix for corner cases)
- Add basic UMAP support
- Add `out/` to `.gitignore` (Visual Studio 2022)

I'm not 100% happy with those table copies but that's for another time to fix.


externals/CMakeExternals.cmake
```diff
@@ -158,6 +158,103 @@ function(require_external NAME)
       GIT_TAG 220fdf37fec2d9d3e3f7674194544ee70eb93ee7 # master on 2021-07-26, because nothing was specified here.
       INCLUDE_DIR "include")
 
+  # aarand
```
moritz-h: The externals should be alphabetically ordered within the header only and reguar library sections.
But I see this is the first case with many libs depdendant on each other. Of they should stay together we should probably think of how they can be wrapped into one depedency or similar.

moritz-h (dismissed):

externals/CMakeExternals.cmake
schulzch: That's not documented anywhere, thus won't fix.

reinago (approved):

PR #1055, opened 2022-06-22
Protein fixes

schatzkn:
Adds a bunch of lost modules ranging from image loaders to the whole molsurfmapcluster plugin

## Summary of Changes

- Adds the molsurfmapcluster plugin again, containing a totally rewritten visualization approach
- Adds image loading capabilities
- Adds the capability to download pdb files directly in the loader


cmake/megamol_config.cmake
```diff
@@ -31,13 +31,30 @@ add_compile_definitions("$<$<CONFIG:DEBUG>:DEBUG>")
...
+# Otherwise, a possible compilation with CUDA will propagate those flags to the CUDA-Compiler and lead to a crash.
+# For certain systems, those flags should be set for both C++ and C compilers to work properly
 if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
   # nothing to do
 elseif (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
-  add_compile_options("-fsized-deallocation") # TODO git history suggests this was required for cuda in 2019, still required?
-  add_compile_options("-Wno-narrowing" "-Wno-non-pod-varargs") # Prevent build fail.
+  # TODO git history suggests this was required for cuda in 2019, still required?
+  add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-fsized-deallocation>) 
+  add_compile_options($<$<COMPILE_LANGUAGE:C>:-fsized-deallocation>)
```
moritz-h: This could be simplified to `add_compile_options($<$<COMPILE_LANGUAGE:C,CXX>:-fsized-deallocation>)`, see https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:COMPILE_LANGUAGE. Same for all other options.

plugins/molecularmaps/CMakeLists.txt
```diff
@@ -23,6 +24,8 @@ if (molecularmaps_PLUGIN_ENABLED)
 
   get_filename_component(cuda_compiler_directory "${CMAKE_CUDA_COMPILER}" DIRECTORY)
 
+  set(CMAKE_VERBOSE_MAKEFILE ON)
```
moritz-h: Looks like a debugging setting. Probably not wanted in master.

plugins/molsurfmapcluster_gl/Shaders/molsurfMapOverlay.btf
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
moritz-h: This will probably fail the Plugin Check GitHub action. `shaders` and `resources` dir must be lowercase. Directory structure must be `shaders/<pluginname>/...`

(And of course btf usage in a fully new plugin should be autorejected xD xD xD)

plugins/imageviewer/include/imageviewer/ImageLoader.h
```diff
@@ -0,0 +1,183 @@
+/*
```
moritz-h: imageviewer2 was renamed to image_gl plugin. Probably we do not want to reintroduce the imageviewer name and merge this stuff here with one of the image plugins. Looks pretty similar from the topic. Also A CMake file für this new plugin seems missing.

moritz-h (changes requested):

plugins/image_gl/src/ImageRenderer.cpp
```diff
@@ -1,15 +1,15 @@
-/**
- * MegaMol
- * Copyright (c) 2010, MegaMol Dev Team
- * All rights reserved.
+/*
```
moritz-h: That actually removes the correct file header with the outdated format

moritz-h (changes requested):

moritz-h (approved):

PR #1062, opened 2022-06-22
PR for osprarrow branch

moritz-h:
PR to archive the `osprarrow` branch.


reinago:
superseded by the way arrows would be rendered in current OSPRay

PR #1084, opened 2022-07-14
Camera Automation for Profiling

gralkapk:


frontend/services/profiling_service/Profiling_Service.cpp
```diff
@@ -52,5 +59,87 @@ void Profiling_Service::resetProvidedResources() {
...
+        "mmGenerateCameraPositions", "(string entrypoint, string camera_path_pattern, uint num_samples)",
+        {[&graph](std::string entrypoint, std::string camera_path_pattern,
+             int num_samples) -> frontend_resources::LuaCallbacksCollection::StringResult {
+            auto entry = graph.FindModule(entrypoint);
+            if (!entry)
+                return frontend_resources::LuaCallbacksCollection::Error{"could not find entrypoint"};
+            auto view = std::dynamic_pointer_cast<core::view::AbstractViewInterface>(entry);
+            if (!view)
+                return frontend_resources::LuaCallbacksCollection::Error{"requested entrypoint is not a view"};
+            auto camera_samples = view->SampleCameraScenes(camera_path_pattern, num_samples);
```
geringsj: codewise, i would argue to really move the `SampleCameraScenes()` outside of the abstract view, somewhere to the utilities, and pass the needed view/camera parameters into that function. even if that means changing the current abstract view interface to pass the necessary bbox. if concerns of views get out of control we soon run into another rewrite towards View_3.

the benefit is that other people can then also sample camera paths (cinematic?), and more importantly you can come up with these camera samples without the need to actually hold a concrete view instance (although i see that since the view has the bbox, this argument is somewhat wonky)

the `camera_path_pattern` could also be a function pointer or closure representing the sampling pattern one wants, instead of making 'resolving the sampling strategy' another problem of the `SampleCameraScenes()`

moritz-h (approved):

PR #1093, opened 2022-07-29
Radical update of mesh_gl plugin

invor:
Removes call/chaining concept for GPU data (meshes, materials, render tasks) in favor of monolithic mesh rendering modules. Still makes use of the existing collection classes for GPU data internally. New base classes are added for mesh renderers to provide a framework for implementing new mesh renderers with as little effort as possible.

## Summary of Changes

- Remove calls for GPU mesh, material and render task data
- Remove abstract base classes for GPU mesh, material and render task modules with chaining
- Add new base classes for monolithic rendering modules 
- Add modules for rendering gltf files with deferred shading and obj files with simple forward shading


plugins/mesh_gl/include/mesh_gl/MeshCalls_gl.h
```diff
@@ -7,71 +7,10 @@
```
moritz-h: Why not just delete that file?

moritz-h (changes requested):

plugins/mesh_gl/include/mesh_gl/MeshCalls_gl.h
invor: Happy?

moritz-h (dismissed):

moritz-h (approved):

PR #1106, opened 2022-08-26
Texture inspector

sellendk:
Introducing a new utility class with which it is possible to view every texture used in a module without explicitly setting the texture as the output. This can be used for debugging. The TextureInspector is based on ImGui and comes with 2 parameter slots. One for selecting the textures the user provides in the constructor of the TextureInspector object and the second for toggling the ImGui window. Additionally, there is a new module TexInspectModule based on the texture inspector. It can be plugged between two modules connected by a CallTexture2D and allows the user to inspect the current texture held by the CallTexture2D.

## Summary of Changes
- new utility class TextureInspector
- added new module TexInspectModule
- removed AntiAliasing ParamSlot 'Output' (replaced by TextureInspector)

## References and Context
The original version of the TextureInspector can be found [here](https://github.com/andyborrell/imgui_tex_inspect).

## Test Instructions
- Include the TextureInspector
- Add a TextureInspector object with a list of texture names (as string vector) in the constructor
- Call the SetTexture() and ShowWindow() methods
- Or, plug the TexInspectModule between two modules connected by a CallTexture2D. Nothing else to do in this case

The [AntiAliasing](https://github.com/sellendk/megamol/blob/19a984397fd8a139afe344e5ddf2ef14c20202a4/plugins/compositing_gl/src/AntiAliasing.cpp#L630-L651) module contains an example


geringsj:
Can you port this code to also support inspection of View outputs, i.e. the contents of the ImageWrapper returned by views when poked for rendering in https://github.com/UniStuttgart-VISUS/megamol/blob/master/frontend/services/image_presentation/ImagePresentation_Service.cpp#L197 ?

@tobiasrau and @invor are also somewhat familiar with the frontend ImageWrapper code.

sellendk:
If I see this correctly, the ImageWrapper holds either a GLTextureHandle or a ByteArray. In case of a GLTextureHandle, the texture inspector can already be used the same way as described above. E. g. in View3DGL the GetRenderingResult() function in the ::Render function returns an ImageWrapper with a GLTextureHandle. You can then do something like this:
```
megamol::frontend_resources::ImageWrapper image = GetRenderingResult();
tex_inspector_.SetTexture(image.referenced_image_handle, image.size.width, image.size.height);
tex_inspector_.ShowWindow();
```
Regarding the ByteArray, as far as I'm aware, there is no built-in function to support bytearrays. So, you would probably need to create a texture from the bytearray and give that handle to the inspector. I can have a look at that, if that's what you were asking.

geringsj:
The usability you describe is what i had in mind for the rendering results of views in the frontend. The ByteArray is not really a problem because there is convenience functionality available to wrap it into a GL texture. Great that this will also be usable in the frontend!

moritz-h (approved):

PR #1112, opened 2022-09-02
Graph Param Subscription Fixes

geringsj:
Fixes broken graph subscription for parameters. 

Parameters not only change their values via lua calls, but also when modules do their normal thing.
For graph subscribers to be able to get notified of _all_ parameter changes (e.g. the GUI needs those), we hook ourselves into the `AbstractParameter.indicateChange()` mechanism and tell the graph about param those changes.

At the very beginning of the next frame, the lua service pokes the graph to broadcast all queued parameter changes to graph subscribers. 

**Entry Point Subscription**
This PR also implements graph subscriptions for entry points. This lets the GUI know when to switch entry points, and thus fixes issues e.g. with SplitView/View2D rendering in the infovis example when switching between views (entry points).

Note that this PR currently does not compile because the implementation of the entry point subscription in the GUI service is intentionally left empty.

### DIFF
--- a/core/src/MegaMolGraph.cpp
+++ b/core/src/MegaMolGraph.cpp
@@ -218,17 +218,46 @@ bool megamol::core::MegaMolGraph::SetParameter(std::string const& paramName, std
     if (!param_ptr)
         return false;
 
+    // unused since generic param changes mechanism does not allow passing 'old value' to graph subscribers
     auto old_value = param_ptr->ValueString();
 
     bool success = param_ptr->ParseValue(value);
 
+    // param changes are queued in the graphs parameter changes queue
+    // and get passed to graph subscribers at beginning of each frame by the Lua Service
+    // using MegaMolGraph::Broadcast_graph_subscribers_parameter_changes()
+
     if (!success)
         return false;
 
+    return true;
+}
+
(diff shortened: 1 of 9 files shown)


core/src/MegaMolGraph.cpp
```diff
@@ -218,17 +218,46 @@ bool megamol::core::MegaMolGraph::SetParameter(std::string const& paramName, std
...
+    return true;
+}
+
+bool megamol::core::MegaMolGraph::Broadcast_graph_subscribers_parameter_changes() {
     for (auto& subscriber : graph_subscribers.subscribers) {
-        if (!subscriber.ParameterChanged(param_slot_ptr, old_value, value)) {
-            log_error("graph subscriber " + subscriber.Name() + " failed to process parameter change: " + paramName +
-                      " from " + old_value + " to " + value);
+
+        for (megamol::core::param::AbstractParamSlot* changed_param_ptr : module_param_changes_queue) {
```
braunms, line 239: At some point module_param_changes_queue has to be cleared.

braunms (approved):

geringsj:
This code should be merged with PR #1111 since the entry point handling is implemented there

PR #1133, opened 2022-11-04
Fix Invisible/Disappearing Parameters in GUI

geringsj:
The new graph subscription mechanism until now only observed actual parameter value changes and passed them to subscribers. This left out the set of parameter state changes regarding GUI visibility, implemented in the AbstractParamPresentation struct and used within the AbstractParam.

This PR moves the AbstractParamPresentation from a base class of the AbstractParam to a member variable, routing previously direct access to GUI presentation state of parameters through the accessor function `AbstractParam.GuiPresentation()` which in turn triggers the internal observer/callback function to notify the graph and subsequently subscribers like the frontend GUI not only of parameter value changes, but also of parameter GUI state changes (e.g. visibility).

`AbstractParam.GuiPresentation()` notifies graph subscribers when the parameter is non-const, while `AbstractParam.GuiPresentation() const` avoids triggering the subscriber callback to prevent a feedback loop inside the GUI code, when the GUI retrieves the GUI presentation state of a parameter. For this, access to the parameter presentation in the GUI is now guarded using `const` variables.

This is a breaking change which forces us to adjust parameter GUI visibility code in the whole code base, like this:
```c++
parameter_ptr->SetGUIVisibility(true); // old
parameter_ptr->GuiPresentation().SetGUIVisibility(true); // new

parameter_ptr->IsGUIVisible(); // old
parameter_ptr->GuiPresentation().IsGUIVisible(); // new
```

## Summary of Changes
* Move AbstractParamPresentation to AbstractParam `gui_presentation` member variable
* Acces to `gui_presentation` through `AbstractParam.GuiPresentation()` triggers graph subscriber callback
* Guard access to param presentation in GUI Service with const
* Search+Replace parameter GUI state setter/getter use in code base according to new access method
* Leave **parameters in probe plugin untouched**, breaking compilation of this plugin

## Test Instructions
Load testspheres.lua and ensure that different render modes of sphere renderer are represented in different visibility of paramters in GUI.


braunms (dismissed):

moritz-h (changes requested):
Maybe think about if the GuiPresentation() indirection is really needed or if it makes things more complicated in the next meeting before merging.

geringsj:
Are there new insights into this?

moritz-h (approved):

PR #1147, opened 2022-11-30
RenderingEndPoint

gralkapk:
Replace SplitView with RenderingEndPoints in ImGui (requires docking feature).

## Summary of Changes

- Infrastructure for passing resources to windows


frontend/services/gui/src/windows/RenderingEndPoint.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+    //entry_point
+    if (current_item != nullptr) {
+        auto ep = img_pres_ep_resource_ptr.get_entry_point(current_item);
+        if (ep.has_value()) {
+            frontend_resources::EntryPoint& ep_v = ep.value();
+
+            ep_v.entry_point_data->update();
+
+            ep_v.execute(ep_v.modulePtr, ep_v.entry_point_resources, ep_v.execution_result_image);
```
geringsj: This is already done by the image presentation service once a framce for each registered entry point. Triggering re-rendering of a frame from inside the GUI seems a bad idea. The intended approach is to then use the `ep_v.execution_result_image` texture handle (either GL or CPU texture) to present the rendered texture in an ImGui window.

geringsj: The rendering execution result image is only available after entry point/graph execution, which introduces a recursive dependency on showing the entry point results in the GUI, which itself is rendered as an entry point.

geringsj: I suppose dragging the frontend resources up until here would also be not necessary without executing the entry points?

frontend/services/gui/src/windows/RenderingEndPoint.h
```diff
@@ -0,0 +1,73 @@
...
+
+#include <map>
+#include <stdint.h>
+
+#include <glad/gl.h>
+
+#include "AbstractWindow.h"
+
+#include "ImagePresentationEntryPoints.h"
+#include "ImagePresentation_Service.hpp"
```
geringsj, line 11: This dependency between GUI (GUI Service) and Image Presentation Service may lead to problems. For cleaner separation of code it will be cleaner to pull the dependencies from `ImagePresentation_Service.hpp` into a separate header or otherwise restructure the affected code.

frontend/services/image_presentation/ImagePresentation_Service.cpp
```diff
@@ -203,18 +220,20 @@ void ImagePresentation_Service::PresentRenderedImages() {
...
+    // pull result images into separate lists
+    std::unordered_map<std::string, std::vector<ImageWrapper>> sink_img_map;
+    sink_img_map.reserve(m_presentation_sinks.size());
 
     // rendering results are presented in order of execution of entry points
     for (auto& entry : m_entry_points) {
-        wrapped_images.push_back(entry.execution_result_image);
+        auto const& sink_list = ep_sink_map[entry.moduleName];
+        for (auto const& sink : sink_list) {
+            sink_img_map[sink].push_back(entry.execution_result_image);
```
geringsj, line 231: I am not sure about the details but it might lead to many GL texture / Byte texture copy operations when putting `ImageWrapper execution_result_image` in different places. I believe the initial goal for the `ImageWrapper` was to have a lightweight abstraction over GL/Byte textures with the option to transform one into the other (texture down/upload). It may be that the `ImageWrapper` copy constructor also copies textures/vectors internally, depending on whether the held texture is only referenced or actually owned.

PR #1148, opened 2022-11-30
GUI Window Resources

gralkapk:
Resource interface for GUI windows.

### DIFF
--- a/frontend/services/gui/src/GUI_Service.hpp
+++ b/frontend/services/gui/src/GUI_Service.hpp
@@ -80,6 +80,7 @@ private:
     Config m_config;
     megamol::core::MegaMolGraph* m_megamol_graph;
     megamol::frontend_resources::PerformanceManager* perf_manager = nullptr;
+    frontend_resources::ProfilingLoggingStatus* perf_logging = nullptr;
     std::shared_ptr<megamol::gui::GUIManager> m_gui = nullptr;
     std::vector<std::string> m_queuedProjectFiles;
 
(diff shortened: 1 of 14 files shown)


frontend/services/gui/src/GUI_Service.hpp
```diff
@@ -84,8 +84,8 @@ class GUI_Service final : public AbstractFrontendService {
     std::vector<std::string> m_queuedProjectFiles;
 
     std::vector<FrontendResource> m_providedResourceReferences;
-    std::vector<FrontendResource> m_requestedResourceReferences;
     std::vector<std::string> m_requestedResourcesNames;
+    std::shared_ptr<frontend_resources::FrontendResourcesMap> frontend_resources;
```
geringsj, line 89: There is no strict need to use a `std::shared_ptr` here. Nothing bad happens if `FrontendResourcesMap` gets copied, everything has value semantics.

gralkapk: The `FrontendResourceMap` carries data that would be copied to every single window though.

geringsj:
The `std::map` contained in the `FrontendResourcesMap` should be fine for copies and the contained resource wrappers are also pretty dumb, the idea beeing that different participants can keep a hard copy of their resource (references) map for convenience, since resources will not change during runtime. This may help maintain isolation of resources such that participants dont (accidentally) get access to resources they did not request. Just trying to give a little context.

reinago (approved):

PR #1150, opened 2023-01-09
Type-safe resource requests

moritz-h:
## Summary of Changes
- Introduce `ResourceRequest` object for type-safe resource requests

### DIFF
--- a/core/src/MegaMolGraph.cpp
+++ b/core/src/MegaMolGraph.cpp
@@ -561,7 +561,7 @@ bool megamol::core::MegaMolGraph::add_module(ModuleInstantiationRequest_t const&
         std::string requested_deps = "";
         std::string found_deps = "";
         for (auto& res : module_resource_request.getResources()) {
-            requested_deps += " " + std::string(res.type.name()); // TODO this is not a printable name!
+            requested_deps += " " + res.TypeName();
         }
         for (auto& dep : module_lifetime_dependencies) {
             found_deps += " " + dep.getIdentifier();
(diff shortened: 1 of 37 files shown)


core/src/MegaMolGraph.cpp
```diff
@@ -550,16 +551,16 @@ bool megamol::core::MegaMolGraph::add_module(ModuleInstantiationRequest_t const&
         return false;
     }
 
-    auto module_lifetime_resource_request = module_ptr->requested_lifetime_resources();
+    frontend_resources::ResourceRequest req;
```
geringsj: I would prefer to use a name like `module_resource_request` instead of `req`

frontend/resources/include/FrontendResourcesLookup.h
```diff
@@ -25,6 +26,41 @@ struct FrontendResourcesLookup {
...
+            if (resource_found) {
+                auto& resource = *dependency_it;
+                result_resources.push_back(request.optional ? resource.toOptional() : resource);
+            } else {
+                if (request.optional) {
+                    result_resources.push_back(megamol::frontend::FrontendResource{}.toOptional());
+                } else {
+                    auto msg = std::string("FrontendResourcesLookup: Fatal Error: ") +
+                               "\n\tcould not find requested resource " +
+                               request.type.name(); // TODO this is not a printable name!
```
geringsj: Maybe wrap `name()` lookup into ResourceRequest helper function

geringsj (approved):

PR #1154, opened 2023-01-16
Static module resource requests

moritz-h:
## Summary of Changes
- Module resource requests are now static

We now could theoretically inject resources already into the module constructor, but injection not changed for now and open for discussion.

### DIFF
--- a/core/src/MegaMolGraph.cpp
+++ b/core/src/MegaMolGraph.cpp
@@ -545,14 +545,8 @@ bool megamol::core::MegaMolGraph::add_module(ModuleInstantiationRequest_t const&
         return false;
     }
 
-    Module::ptr_type module_ptr = module_description->CreateModule(request.id);
-    if (!module_ptr) {
-        log_error("error. could not instantiate module from module description: " + request.className);
-        return false;
-    }
-
     frontend_resources::ResourceRequest module_resource_request;
-    module_ptr->requested_lifetime_resources(module_resource_request);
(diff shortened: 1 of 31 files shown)


core/src/MegaMolGraph.cpp
```diff
@@ -574,6 +568,12 @@ bool megamol::core::MegaMolGraph::add_module(ModuleInstantiationRequest_t const&
         return false;
     }
 
+    Module::ptr_type module_ptr = module_description->CreateModule(request.id);
```
geringsj, line 571:
```suggestion
    Module::ptr_type module_ptr = module_description->CreateModule(request.id, module_lifetime_dependencies);
```

I guess the next step is to pass the actual resources through the CreateModule() infrastructure to the module constructors, right?

moritz-h: I'm not sure if we want to do this. First, with the current module concept the constructors should be as lightweight as possible and basically do nothing else except slot registration to be dummy constructible by the configurator. I don't know if providing resources in the constructor is needed therefore. Second, it would be a massive breaking change, needing to touch every single module constructor. (Basically the same argument as for the idea of putting the resource request in the constructor).
Therefore, I would currently vote for keepings things as they are, at least until we finished thinking about the whole core refactoring. With making the resource request static we at least removed them from being bound to the module runtime and have the flexibility to inject dependencies into the constructor in future.

geringsj (approved):

Issue #1114, opened 2022-09-06
Default value for "host" is unreasonable


crowbar27:
**Describe the bug**
The default value for `mmSetCliOption("host", "tcp://*:33333")` is the loopback adapter. It is unreasonable that the *remote* console is bound to the loopback adapter by default.

**To Reproduce**
1. Start MegaMol
2. Run `netstat -ano`
3. See that the console is bound to 127.0.0.1:3333

**Expected behavior**
The remote console should be bound to the any address in order to allow for remote access.

**Environment**
 - Windows 10
 - NVIDIA GeForce RTX 3090 Ti

moritz-h:
With #1157 now at least the host config is within the default config file and could be easily edited. Changing the default to accept any remote connection is probably very bad from security perspectives. At least until somebody implements authentication on the remote console.

PR #1157, opened 2023-01-24
Network refactoring for remoteconsole

moritz-h:
## Summary of Changes
- Use a single zmq socket for all connections (ZMQ_ROUTER) instead of the previously used two step connection startup. This simplifies the code, removes unnecessary threads and makes MegaMol a lot more firewall friendly, because no additional randomly chosen ports are used.


geringsj:
While we are at it, we could also fix Issue https://github.com/UniStuttgart-VISUS/megamol/issues/1114 and change the default port for host to something other than `33333`

moritz-h:
#1114 is not about changing the port, but changing the network socket to listen on any incoming connection instead of just on localhost. So setting host address to `tcp://*:33333` instead of `tcp://127.0.0.1:33333`. This would make it way more convenient when the remote console just works. On the other hand this is a probably a pretty bad default state from security perspective. Opening a remote port, accepting any incoming connection (without any form of authentication) and allowing arbitrary lua code execution is probably not far from a security nightmare.
Maybe instead we just add the setting to the default config file, that you can just edit the config directly instead of must finding out the correct command every time. A little more work would be to add some basic password authentication to the remote console or some other security solution.
Maybe we should bring this up in the next team meeting as it is the classical convenience vs security design decision.

frontend/services/lua_service_wrapper/Lua_Service_Wrapper.hpp
```diff
@@ -84,8 +80,7 @@ class Lua_Service_Wrapper final : public AbstractFrontendService {
 
     int m_service_recursion_depth = 0;
 
-    // auto-deleted opaque ZMQ networking object from LuaHostService.h
-    std::unique_ptr<void, std::function<void(void*)>> m_network_host_pimpl;
+    std::unique_ptr<megamol::frontend::LuaRemoteConnectionsBroker> m_network_host;
```
geringsj, line 83: I guess if we directly use the `LuaRemoteConnectionsBroker` type we also dont need the `std::unique_ptr` anymore?

geringsj (approved):

PR #1170, opened 2023-03-11
CSVFileSequence: access data source module via call

geringsj:
This PR patches the CSVFileSequence to access a connected data module (via `inSlot`) to modify the target filename param using the call connected to `inSlot`, instead of using the MegaMolGraph resource to access the data module globally.

This remove the MegaMolGraph as a resource requirement from CSVFileSequence and thus future proofs the module for a time when the graph can not be accessed or manipulated using that resource.

## Summary of Changes
* modify CSVFileSequence target filename param access
* remove MegaMolGraph resource access
* small fixes in FlagStorage and TableColumnFilter
* AbstractNamedObject fix prepending too many :: to module names
* other minor fixes

## Test Instructions
`csv\csv_file_sequence.lua` in the examples repository, along with `sampledata\csv_sequence\testspheres.*.csv` as file sequence inputs:

https://user-images.githubusercontent.com/44215465/230109790-b8ff2693-298c-4f49-97b2-1176c841f224.mp4

### DIFF
--- a/core/src/LuaAPI.cpp
+++ b/core/src/LuaAPI.cpp
@@ -1,7 +1,7 @@
 /*
- * LuaState.cpp
+ * LuaAPI.cpp
  *
- * Copyright (C) 2017 by Universitaet Stuttgart (VIS).
+ * Copyright (C) 2020 by Universitaet Stuttgart (VIS).
  * Alle Rechte vorbehalten.
  */
 
--- a/core/src/AbstractNamedObject.cpp
+++ b/core/src/AbstractNamedObject.cpp
@@ -36,7 +36,8 @@ vislib::StringA AbstractNamedObject::FullName() const {
                 break;
             }
             name.Prepend(ano->Name());
-            name.Prepend("::");
+            if (!name.StartsWith("::", false))
+                name.Prepend("::");
             ano = ano->Parent();
         }
         return name;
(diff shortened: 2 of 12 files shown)


core/src/LuaAPI.cpp
```diff
@@ -1,7 +1,7 @@
 /*
- * LuaState.cpp
+ * LuaAPI.cpp
  *
- * Copyright (C) 2017 by Universitaet Stuttgart (VIS).
+ * Copyright (C) 2020 by Universitaet Stuttgart (VIS).
```
invor, line 4: I think we discussed a while ago not to update the initial year of copyright

plugins/mmstd/include/mmstd/flags/FlagStorage.h
```diff
@@ -122,7 +122,10 @@ class FlagStorage : public core::Module {
     core::param::ParamSlot serializedFlags;
 
     std::shared_ptr<FlagCollection_CPU> theCPUData;
-    uint32_t version = 0;
+
+    // set initial version (in first frame, with temp data) != 0 so call data comes out as "has Update"
+    // and eventually triggers a reload of actual user-provided data via validateFlagCount()
+    uint32_t version = 1;
```
invor: Is there a specific module that requires this or a general fix? Should validateFlagCount() maybe be called independent from hasUpdate? E.g. when changing data (size) this will also not be reflected by an update inside the flag storage and instead connected modules need to inform the flag storage about the change in size.

invor (commented):
Are the example projects already uploaded?

geringsj:
Example project is at: https://github.com/UniStuttgart-VISUS/megamol-examples/pull/14

plugins/mmstd/include/mmstd/flags/FlagStorage.h
geringsj: I think initially i fixed this for the SphereRenderer+PCP use case, where time dependent particle data did not get updated via the FlagStorage.

moritz-h (approved):

PR #1177, opened 2023-04-13
Shader validation

moritz-h:
## Summary of Changes
- Add automated shader validation using glslang


geringsj (dismissed):
I guess this works and LGTM. Do all our shaders currently have .glsl file endings, along with the appropriate shader stage annotations? Wondering whether we cover all shaders.

moritz-h:
> Do all our shaders currently have .glsl file endings, along with the appropriate shader stage annotations?

That all shaders must have a .glsl extension is a style guide decision, see https://github.com/UniStuttgart-VISUS/megamol/blob/master/docs/devguide.md#shader-files. The shader stage before .glsl is required by the shader factory to determine the type. So a missing shader stage would also not work at runtime (side note, it would technically work with just the stage as extension without the .glsl).
Actually, this is practically used here, because we can only validate full shaders and not snippets independently with glslang. All shader snippets do not have that stage before .glsl. So the format *.<stage>.glsl is used to determine what should be passed to the validation.

geringsj:
> > Do all our shaders currently have .glsl file endings, along with the appropriate shader stage annotations?
> 
> That all shaders must have a .glsl extension is a style guide decision

Sounds like someone thought this through :thumbsup:

moritz-h:
Now with CI check for the shader file extensions to enforce style guidelines :smile:

geringsj (approved):

PR #1184, opened 2023-04-19
Fixed bug for unwanted call curve hovering in configurator

braunms:
## Summary of Changes
- prevent call curve hovering when a pop-up is open
- fixed profiling button position of modules
- allow moving module pop-up


moritz-h (approved):
Tested. LGTM.

PR #1192, opened 2023-05-24
ClipPlane GUI Widget

braunms:
Added the imguizo library to provide an interactive manipulator widget for the clipping plane module.

The ImGui part is realized as parameter group widget in the GUI frontend. 
Callers have to send the current camera first, before they request the clip plane via the callback of the ClipPlaneCall.

TODO
- [ ] Rename ClipPlane module to PlaneModule?
- [ ] Rename ClipPlaneCall call to PlaneCall?

ISSUES
- [ ] Widget gets clipped at internal near plane that is way too near....

FEATURES
- [ ] button to invert the normal of the plane to switch between clipping "in front" and "behind"
- [ ] separate between "enabled" (does some clipping) and "manipulatable" (widgets are shown and can be moved); maybe add option under draw to enable/disable drawing the dragable widgets

Already contains changes of PR #1195 and #1194 ...

## Test Instructions

- Already implemented usage: See changes in SphereRenderer ...
- Add the following code to your renderer in Render():

```C++
    // Requires call slot 'get_clip_plane_slot_':
    // this->get_clip_plane_slot_.SetCompatibleCall<view::CallClipPlaneDescription>();
    // this->MakeSlotAvailable(&this->get_clip_plane_slot_); 
    megamol::core::view::Camera cam = call.GetCamera();
    view::CallClipPlane* ccp = this->get_clip_plane_slot_.CallAs<view::CallClipPlane>();
    if (ccp != nullptr) {
        ccp->SetCamera(cam);
    }
    if ((ccp != nullptr) && (*ccp)()) {
        auto plane = ccp->GetPlane();
    }
```


straubar:
Code looks fine so far. Just found a spelling error for one of the variables: `scree_pos` instead of `screen_pos` in `ParameterGroupClipPlaneWidget.[h|cpp]`.

I will look at the functionality later and give it a test run.

straubar:
There are still some issues:
- camera and plane position influences the clipping (?) of the dragable widgets of the clip plane (cf. image)
- dragging the plane outside of the bounding box leads to widgets disappearing, and jumping of the clip plane when then moving the camera
- direct modification of the x-component of the clip plane's point leads to weird update, resulting in a different value than the one entered (this only happens after having dragged the plane outside of the bounding box and after the jump occurred, as mentioned above; may disappear if above is fixed)
- the rotation does not seem intuitive (to me). is this as intended?
- widgets for manipulation should not be visible if the clip plane is disabled (see also suggestion below)

feature suggestion:
- button to invert the normal of the plane to switch between clipping "in front" and "behind"
- separate between "enabled" (does some clipping) and "manipulatable" (widgets are shown and can be moved); maybe add option under draw to enable/disable drawing the dragable widgets

grafik

straubar:
Lua project for above comments:
[test-clip.zip](https://github.com/UniStuttgart-VISUS/megamol/files/11678518/test-clip.zip)

PR #1194, opened 2023-06-01
Configurator: Fix additional "ghost" module when adding new module to group

braunms:
If a new module is created inside a group in the Configurator, the looped back notification from the MegaMol graph for adding the new module can not be discarded by the GUI and a new "ghost" module was added. The reason is that in the GUI graph the module is already renamed and added to a group and therefore has a different name.

**Since the origin of the notifications from the MegaMol graph can not be detected, the GUI graph should only add ONE queue event per module per frame!**

## Summary of Changes
- Adapted code that group name of newly added module is known before the module is added. Therefore, the subsequent renaming can be prevented.
- Cleaned up redundant code for adding and removing modules to a group.


reinago (approved):
did a bit of testing, looked good so far

PR #1205, opened 2023-07-04
Highlighting of Parameters in the GUI

gralkapk:
Adds new Lua Callback that allows to mark a parameter in the GUI.

## Summary of Changes

- mmSetParamHighlight
- maybe more?

### DIFF
--- a/core/src/MegaMolGraph.cpp
+++ b/core/src/MegaMolGraph.cpp
@@ -222,6 +222,17 @@ bool megamol::core::MegaMolGraph::SetParameter(std::string const& paramName, std
     return true;
 }
 
+bool megamol::core::MegaMolGraph::SetParameterHighlight(std::string const& paramName, bool is_highlight) {
+    auto param_slot_ptr = FindParameterSlot(paramName);
+    auto param_ptr = getParameterFromParamSlot(param_slot_ptr);
+
+    if (!param_ptr)
+        return false;
+    param_ptr->SetGUIHighlight(is_highlight);
+
+    return true;
+}
+
 bool megamol::core::MegaMolGraph::Broadcast_graph_subscribers_parameter_changes() {
     for (auto& subscriber : graph_subscribers.subscribers) {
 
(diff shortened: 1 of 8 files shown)


core/src/MegaMolGraph.cpp
```diff
@@ -222,6 +222,17 @@ bool megamol::core::MegaMolGraph::SetParameter(std::string const& paramName, std
...
+    auto param_slot_ptr = FindParameterSlot(paramName);
+    auto param_ptr = getParameterFromParamSlot(param_slot_ptr);
+
+    if (!param_ptr)
+        return false;
+    param_ptr->SetGUIHighlight(is_highlight);
+
+    return true;
+}
+
```
geringsj: I prefer we dont pollute the graph API with hacks for unrelated features. The graph should support only a minimal set of operations. 

It is totally fine to simply get a parameter from the graph and ask it directly to do things, e.g:

https://github.com/UniStuttgart-VISUS/megamol/blob/master/frontend/services/lua_service_wrapper/Lua_Service_Wrapper.cpp#L567-L575

https://github.com/UniStuttgart-VISUS/megamol/blob/master/frontend/services/lua_service_wrapper/Lua_Service_Wrapper.cpp#L580-L587

core/src/param/AbstractParamPresentation.cpp
```diff
@@ -203,6 +205,10 @@ bool AbstractParamPresentation::StateFromJSON(const nlohmann::json& in_json, con
                         valid &=
                             megamol::core::utility::get_json_value<bool>(gui_state, {"gui_read-only"}, &gui_read_only);
 
+                        bool gui_highlight = false;
+                        valid &=
+                            megamol::core::utility::get_json_value<bool>(gui_state, {"gui_highlicht"}, &gui_highlight);
```
braunms: Typo, json key should be "gui_highlight"

geringsj (approved):

PR #1213, opened 2023-07-18
Porous Media Flow Graph

straubar:
Implementation for the paper "Visual Analysis of Displacement Processes in Porous Media using Spatio-Temporal Flow Graphs".


straubar:
Contains #1207, which should be merged first.

reinago (approved):

PR #1232, opened 2023-10-12
Update clang-format

moritz-h:
## Summary of Changes
- Update clang-format to version 16 (current VS default)
- Update deprecated rules, min version of rules file is now 14
- We can now switch to `SpaceAfterCStyleCast: true`, we initially used false because of inconsistent behavior of clang-format-12


geringsj (approved):
LGTM

